PR #4, opened 2020-06-22
Reformat with black

codeinthehole:


Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.7.7-slim
```
codeinthehole, line 1: Is it a good practice to use `-slim` images?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Create virtualenv and add to path.
+ENV VIRTUAL_ENV=/opt/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
codeinthehole, line 6: Is it a good practice to use a virtualenv within a Docker container?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# System dependencies
+RUN apt-get update && apt-get install -y build-essential && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /opt/app
```
codeinthehole, line 11: Is it possible to mount relative to this `WORKDIR`? It feels clunky to have to specify it again when running tests:
```
$ docker run --rm -v `pwd`:/opt/app xocto
```

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.7.7-slim
+
+# Create virtualenv and add to path.
+ENV VIRTUAL_ENV=/opt/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# System dependencies
+RUN apt-get update && apt-get install -y build-essential && rm -rf /var/lib/apt/lists/*
```
codeinthehole, line 9: Once these system dependencies are installed, should I create a non-privileged user and run everything else as that?

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Install Python requirements. README.md is required as the setup.py file
+# refers to it.
+COPY README.md .
+COPY setup.py .
+RUN pip install -e .[dev,test]
+
+# Run py.test against current dir by default but allow custom args to be passed
+# in.
+ENTRYPOINT ["py.test"]
+CMD [""]
```
codeinthehole, line 22: This makes it easy to run tests but a bit of a pain to run, say, `isort`. You have to do this:
```bash
$ docker run --rm -v `pwd`:/opt/app --entrypoint=isort xocto -rc
```

Better to have a more general-purpose run-something-within-the-venv image rather than a specialist run-tests-within-the-venv?

Dockerfile
scalen: I've not seen it done, and I can't find a reference to it online, I'm afraid :-( would be nice though!

Dockerfile
dooferlad: Don't see the point - everything is root anyway.

Dockerfile
dooferlad: no - the container is isolated from the rest of your system anyway.

Dockerfile
dooferlad: I may get into trouble for saying this, but I think the Docker way of doing things is quite clear - they expect you to install everything using a Dockerfile and have a single command that the container runs and that is it. It isn't designed to run different commands, so if you want to do that then it will feel painful and I think that is a design decision on their part. 

Using docker to run all the other parts of your dev environment (databases etc) is a great idea, but changing the code inside a container just to get the Python environment you want? Just use a virtualenv!

All of that said, _if_ you want to use Docker like this then you can use `docker exec` because you don't need a new container (exec = existing container, run = new container).

scalen: That depends on how you see this being used... I would usually say that an image meant to run a program should have an entrypoint, one meant to contain a dev environment should not.

If you're looking to make a single image run multiple programs, you're doing it kinda wrong :-P I would suggest not specifying the entrypoint or command in the base image, then extend that image in a series of named later stages:
```
FROM python:3.7.7-slim as base

...

FROM base as pytest

# Run py.test against current dir by default but allow custom args to be passed
# in.
ENTRYPOINT ["py.test"]
CMD [""]

FROM base as black

# Run black against current dir by default but allow custom args to be passed
# in.
ENTRYPOINT ["black"]
CMD [""]
```

You can then build these using `docker build -t xocto/pytest --target pytest . && docker build -t xocto/black --target black .`, and basically the whole image should be shared.

README.md
```diff
@@ -58,14 +58,22 @@ Create and activate a virtualenv then:
...
 
     $ make test
 
 and:
 
     $ make lint  
 
+A development docker image can be built with:
+
+    $ docker build . -t xocto
```
scalen, line 71: see my previous comments.

README.md
```diff
@@ -58,14 +58,22 @@ Create and activate a virtualenv then:
...
+
+and tests can be run within this container with:
+
+    $ docker run -v `pwd`:/opt/app xocto
```
scalen, line 75: Based on my previous comments:
```suggestion
    $ docker run -v `pwd`:/opt/app xocto/pytest

Black can also be run as follows:

    $ docker run -v `pwd`:/opt/app -w /opt/app/src xocto/black
```

Dockerfile
scalen: @dooferlad This isn't quite true, as the expected behaviour is to mount host files to the container.  Leaving the user as root in this case give the container root access to all host files in the mounted directory, as the root user id is always 1.

On the other hand, any user you create in this dockerfile cannot be assumed to match any existing user on the host, so this will mean the container will either have full root access or absolute minimum access to the mounted files.

Dockerfile
scalen: @dooferlad You are definitely right about the docker images in general.  However, I do kinda see the point of having a perfectly sandboxed and consistent testing/linting environment that you just inject your code into.  This avoids things like global linting configuration accidentally impacting your linting for a tightly controlled individual project.  So I can see the point.

Dockerfile
codeinthehole: Some light googling suggests there is a point when doing multi-stage builds as it's easier to copy over just the site-packages that your application uses: https://pmac.io/2019/02/multi-stage-dockerfile-and-python-virtualenv/

Dockerfile
codeinthehole: Good to know

Dockerfile
codeinthehole: Thanks both - that's interesting. It doesn't really matter for this project - I was just using it as a sandbox.

I like @scalen's idea of creating multiple single-process images from the same base. I find it a bit weird that building the `black` image still runs all the commands from the `pytest` image but I guess it does not harm when the leaf images just contain `ENTRYPOINT` and `CMD` commands (which don't create new layers).

codeinthehole: > but changing the code inside a container just to get the Python environment you want? Just use a virtualenv!

Having a dev image like this might make sense when building the virtualenv is heavy-weight (ie lots of compilation). I thought that was the case here due to pandas but later realised I could upgrade and install a wheel instead.

scalen: > I find it a bit weird that building the `black` image still runs all the commands from the `pytest` image but I guess it does not harm when the leaf images just contain `ENTRYPOINT` and `CMD` commands (which don't create new layers).

It's true that it doesn't matter much as long as the pytest image is on the same box, but you're right, it could be smaller and still based on the same stack of layers.  You'd just need to break up the commands in the `base` image into smaller stages, and base the `black` image off a stage before the layers relating to `pytest` 🤷

Dockerfile
scalen: That is true, but you aren't copying the files between stages in this Dockerfile :-P I plan on moving the kraken-core image to this at some point, once I work out how to copy from a venv to the global python modules.

PR #5, opened 2022-04-14
Migration from kraken core  - Stage 1/?

jbkkd:


codeinthehole (dismissed):
No objection in principle.

Of course, we need to remember `xocto` is open source (at least until we create a private PyPI registry).

Can you bring each module's tests across too? And update the docs a bit (i.e. add some examples to the README).

jbkkd:
Will do on both.

I also plan on duplicating the rules we have for isort and black so they follow the same conventions (they're slightly different on this repo). Still want to be able to extract them out to a shared "linting" library or something similar.

jbkkd:
Needs to merge before https://github.com/octoenergy/kraken-core/pull/51291

ewanjones (approved):
LGTM

PR #7, opened 2022-04-28
Define text/markdown as the long description content type

jbkkd:
See https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ setup(
     version=VERSION,
     description="Octopus Energy Python service utilities",
     long_description=long_description,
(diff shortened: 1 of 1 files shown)


samueljsb (approved):

ahter:
@jbkkd have you see the linting task error in CircleCI?
```
AttributeError: 'FlakesChecker' object has no attribute 'CONSTANT'
```

Is that expected?

jbkkd:
@ahter woops - I completely missed that as the build itself did pass. I'll look into it

PR #14, opened 2022-06-07
Add get_month_datetime_ranges function

bendamschke:


tests/test_localtime.py
```diff
@@ -1023,3 +1023,71 @@ def test_timestamp_british_summer_time_after_clocks_move_backward(self):
         assert dt.day == 25
         assert dt.hour == 1
         assert dt.minute == 30
+
+
+class TestGetMonthDatetimeRanges:
+    def test_yields_correct_ranges(self):
```
leamingrad, line 1029: 🤔 This test is good, but I think it would be helpful to check a couple of other scenarios. Specifically:
* What happens if the input range is not aligned to month boundaries?
* What happens if the timezone is a weird one (a normal-weird example would be AUS, a super-weird one would be Samoa between 2011 and 2012)

xocto/localtime.py
```diff
@@ -664,3 +664,22 @@ def translate_english_month_to_spanish(month: int) -> str:
         "December": "deciembre",
     }
     return month_name_lookup[month_name]
+
+
+def get_month_datetime_ranges(
+    range: ranges.FiniteDatetimeRange,
```
leamingrad, line 670: 🤔 To match the other functions in this file, could this function be modified to take an optional `tz` argument, so that the timezone can be specified?

Alternatively, we should probably assert that the input range boundaries are localised in some way.

leamingrad (changes requested):
Thanks for this, but I think we need to be a bit more careful around the handling of timezones in `localtime`.

tests/test_localtime.py
leamingrad: > Do you mean providing the timezone within the datetime objects, or as an argument for the function? 

Both, but the latter is the most important

PR #15, opened 2022-07-29
Use Python's datetime UTC instead of pytz

jamesbeith:
This change swaps uses of pytz UTC to instead use Python's datetime UTC as part of an effort to remove the legacy pytz library altogether.

Merging this PR will incur a major version bump of the xocto package.

### DIFF
--- a/tests/test_localtime.py
+++ b/tests/test_localtime.py
@@ -854,6 +854,12 @@ class TestCombine:
                 datetime.datetime(2020, 1, 1, tzinfo=pytz.UTC),
             ),
             (
+                factories.date("1 Jan 2020"),
+                factories.time("00:00"),
+                "UTC",
+                datetime.datetime(2020, 1, 1, tzinfo=datetime.timezone.utc),
+            ),
+            (
                 factories.date("1 Jun 2020"),
                 factories.time("01:00"),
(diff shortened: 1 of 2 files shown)


xocto/localtime.py
```diff
@@ -278,6 +278,8 @@ def latest(_date=None, tz=None):
 
 def combine(_date: datetime_.date, _time: datetime_.time, tz) -> datetime_.datetime:
     combined_dt = datetime_.datetime.combine(_date, _time)
+    if tz is datetime_.timezone.utc:
+        return combined_dt.replace(tzinfo=tz)
```
jamesbeith, line 282: We need this because the `localize` method only exists on a pytz timezone.

tests/test_localtime.py
```diff
@@ -853,6 +853,12 @@ class TestCombine:
                 "UTC",
                 datetime.datetime(2020, 1, 1, tzinfo=pytz.UTC),
             ),
+            (
```
seddonym, line 856: 🐼 Instead of adding an extra parameter, we could add a second parametrize call that uses either `pytz.UTC` or `datetime.timezone.utc` for each of the existing parameters.

seddonym (approved):

PR #16, opened 2022-10-14
Remove the `assert_never` helper

samueljsb:
This function is now part of the stdlib `typing` module from Python 3.11
and has been backported in `typing_extensions`. A custom implementation
is no longer necessary now there is a standard implementation.

---

This is technically a breaking change. Would it be better to continue to export this function but just proxy the one from `typing`/`typing-extensions`?

### DIFF
--- a/xocto/types.py
+++ b/xocto/types.py
@@ -2,7 +2,7 @@
 Utility types to save having to redefine the same things over and over.
 """
 
-from typing import Generic, NoReturn, Protocol, Tuple, TypeVar, Union
+from typing import Generic, Protocol, Tuple, TypeVar, Union
 
 from django.contrib.auth import models as auth_models
 from django.db import models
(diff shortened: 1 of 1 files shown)


leamingrad (approved):
🚢 I think this make sense - we can just update the refs next time we update xocto

leamingrad:
I think that the coconut team have their own implementation, but otherwise nothing in the organisation uses it: https://github.com/search?l=&p=5&q=assert_never+user%3Aoctoenergy&ref=advsearch&type=Code

It might be worth checking with @jbkkd though as he probably knows the most about where xocto is used.

jbkkd:
Outside of `kraken-core`, `xocto` is mainly used in `octotrade`, though not the `types` module, so you're good to go

jbkkd:
And also this is a public repo, I don't know that we have many users (if at all) as we haven't documented this repo very well.
One for a future spa day.

PR #17, opened 2022-10-19
Fix CI

samueljsb:
CI hasn't been running for a few reasons. This fixes them.


cloverlime (approved):
LGTM, but why do the tests require a Django project?

samueljsb:
They require a Django project because one of the modules imports `django.conf.settings` and tries to read from it.

PR #25, opened 2023-01-13
Setup autodoc

pydanny:
- Loaded Django into Sphinx so autodoc would work
- Added more directory levels to the path so autodoc can discover Python modules
- Added autodoc for all modules

Note: The Sphinx theme we're using isn't the best for autodoc, we'll move to a new theme in a separate PR.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,15 +10,16 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
+import os
+import sys
 
+import django
 
-# TODO: Use with autodoc
-# from django.conf import settings
+sys.path.insert(0, os.path.abspath(".."))
+sys.path.insert(0, os.path.abspath("."))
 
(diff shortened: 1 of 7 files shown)


pydanny (commented):
Ugh, tried to resolve a merge conflict quickly and created a merge commit 🤦‍♂️

jbkkd:
Just do an interactive rebase and the merge commit will disappear

pydanny:
Heh, I'm so out of practice with team-based git usage. I keep compounding my errors. 🔥

pydanny:
Okay, commit issue is fixed now. It will be good to get my git-fu back up.

docs/conf.py
```diff
@@ -10,15 +10,16 @@
...
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
+import os
+import sys
 
+import django
 
-# TODO: Use with autodoc
-# from django.conf import settings
+sys.path.insert(0, os.path.abspath(".."))
```
jbkkd, line 18: Could use a comment on why we're inserting both this folder and the one above it to the path

jbkkd (approved):
Can relate - I've been in the world of "How do I code?" recently.

PR looks good!

PR #27, opened 2023-01-17
Add get last modified function

aplusclients:
The purpose of this PR is to add a function that would allow easier access when needing to simple check the last modified date for an s3 file.

### DIFF
--- a/tests/storage/test_storage.py
+++ b/tests/storage/test_storage.py
@@ -1,5 +1,6 @@
 import builtins
 import os
+import datetime
 import shutil
 import tempfile
 from unittest import mock
@@ -747,6 +748,29 @@ class TestLocalFileStore:
         store = storage.LocalFileStore("bucket-name")
         fetch_url = store.fetch_url("some/key", version_id="some-version")
         assert fetch_url == "/media-url/bucket-name/some/some-version/key"
+
+
+    @mock.patch.object(storage.S3FileStore, "_get_boto_object")
+    def test_get_last_modified_given_path(self, get_boto_object):
+        """Fetch file should act on the given path.
+
+        It should not add the subdirectory to the start of the key.
+        """
+        bucket_name = "some-bucket"
+        directory = "folder"
+        store = storage.S3SubdirectoryFileStore(f"s3://{bucket_name}/{directory}")
+        base_key = "my/file.txt"
+
+        _last_modified = get_boto_object().last_modified
+
+        full_key = store.get_key(base_key).key
+        assert full_key == os.path.join(directory, base_key)
+
+        s3object = storage.S3Object(bucket_name, full_key)
+
+        store.get_last_modified(s3_object=s3object)
+
+        assert store.get_last_modified(s3_object=s3object) == _last_modified
 
 
 class TestFromUri:
--- a/tests/storage/test_storage.py
+++ b/tests/storage/test_storage.py
@@ -1,6 +1,5 @@
 import builtins
 import os
-import datetime
 import shutil
 import tempfile
 from unittest import mock
@@ -748,7 +747,6 @@ class TestLocalFileStore:
         store = storage.LocalFileStore("bucket-name")
         fetch_url = store.fetch_url("some/key", version_id="some-version")
         assert fetch_url == "/media-url/bucket-name/some/some-version/key"
-
 
     @mock.patch.object(storage.S3FileStore, "_get_boto_object")
     def test_get_last_modified_given_path(self, get_boto_object):
--- a/xocto/storage/storage.py
+++ b/xocto/storage/storage.py
@@ -701,7 +701,7 @@ class S3FileStore(BaseS3FileStore):
         boto_object = self._get_boto_object(s3_object=s3_object)
         return boto_object.content_length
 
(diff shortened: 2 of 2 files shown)


xocto/storage/storage.py
```diff
@@ -1179,6 +1184,11 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
         file_stats = os.stat(filepath)
         return file_stats.st_size
 
+    def get_last_modified(self, *, s3_object: S3Object) -> datetime:
```
davetran55, line 1187: the return type should be `datetime.datetime`

xocto/storage/storage.py
```diff
@@ -1179,6 +1184,11 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
...
+        filepath = os.path.join(self.storage_root, s3_object.bucket_name, s3_object.key)
+        file_stats = os.stat(filepath)
+        return datetime.datetime(file_stats.st_mtime)
```
davetran55, line 1190: I think it should be something like this `return datetime.datetime.fromtimestamp(file_stats.st_mtime)`

xocto/storage/storage.py
```diff
@@ -700,6 +701,10 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
         boto_object = self._get_boto_object(s3_object=s3_object)
         return boto_object.content_length
 
+    def get_last_modified(self, *, s3_object: S3Object) -> datetime:
```
davetran55, line 704: the return type should be datetime.datetime

davetran55 (commented):
Provided some comments.  Can we add couple of tests here?

xocto/storage/storage.py
aplusclients: Updated, thank you.

tests/storage/test_storage.py
```diff
@@ -748,6 +748,28 @@ def test_fetch_url_with_version(self):
...
+        _last_modified = get_boto_object().last_modified
+
+        full_key = store.get_key(base_key).key
+        assert full_key == os.path.join(directory, base_key)
+
+        s3object = storage.S3Object(bucket_name, full_key)
+
+        store.get_last_modified(s3_object=s3object)
+
+        assert store.get_last_modified(s3_object=s3object) == _last_modified
```
davetran55, line 771: Can you add this assertion to make sure we return a `datetime.datetime`?
`assert type(store.get_last_modified(s3_object=s3object)) == datetime.datetime`

xocto/storage/storage.py
```diff
@@ -1179,6 +1184,10 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
         file_stats = os.stat(filepath)
         return file_stats.st_size
 
+    def get_last_modified(self, *, s3_object: S3Object) -> datetime.datetime:
```
davetran55, line 1187: Because you are in local file system, I don't think this will work.  Your implementation in the prior commit should work.
```
    def get_last_modified(self, *, s3_object: S3Object) -> datetime.datetime:
        filepath = os.path.join(self.storage_root, s3_object.bucket_name, s3_object.key)
        file_stats = os.stat(filepath)
        return datetime.datetime.fromtimestamp(file_stats.st_mtime)
```

xocto/storage/storage.py
```diff
@@ -1179,6 +1184,11 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
         file_stats = os.stat(filepath)
         return file_stats.st_size
 
+    def get_last_modified(self, *, s3_object: S3Object) -> datetime:
+        filepath = os.path.join(self.storage_root, s3_object.bucket_name, s3_object.key)
+        file_stats = os.stat(filepath)
```
durden: ❓ Are we sure that `os.stat()` works properly on a file from S3?

xocto/storage/storage.py
```diff
@@ -701,7 +701,7 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
         boto_object = self._get_boto_object(s3_object=s3_object)
         return boto_object.content_length
 
-    def get_last_modified(self, *, s3_object: S3Object) -> datetime:
+    def get_last_modified(self, *, s3_object: S3Object) -> datetime.datetime:
```
durden, line 704: This should be in a fixup commit and rebased into the commit that added this to avoid non-atomic commits where the type checker is broken.

tests/storage/test_storage.py
durden: ❓ Isn't that exactly what `mypy` is for?

I'm not exactly sure what this test is doing It gets the last modified date from the object being tested, calls a method then checks that it returns what it said?

This feels like it's testing that `mock.patch.object` works not the code under test. I feel like we should be _setting_ something in `last_modifed` and make sure it returns what we expect. The way this is written now it seems like it:
- Asks the boto object what it's going to return for last modified
- Calls the boto object last modified
- Verify it returns what it said it would

What does this actually test?

aplusclients:
Completed by @davetran55 with this PR https://github.com/octoenergy/xocto/pull/32

PR #32, opened 2023-01-19
Add last modified

davetran55:
This PR adds last modified fields to S3FileStore, MemoryFileStore and LocalFileStore

### DIFF
--- a/tests/storage/test_storage.py
+++ b/tests/storage/test_storage.py
@@ -440,7 +440,7 @@ class TestS3FileStore:
     @mock.patch.object(storage.S3FileStore, "_get_boto_object_for_key")
     def test_get_last_modified(self, get_boto_object_for_key):
         k = mock.Mock()
-        k.last_modified = datetime.datetime(2023,1,1,8,0,0,0)
+        k.last_modified = datetime.datetime(2023, 1, 1, 8, 0, 0, 0)
         get_boto_object_for_key.return_value = k
         store = storage.S3FileStore("bucket")
 
--- a/tests/storage/test_storage.py
+++ b/tests/storage/test_storage.py
@@ -440,7 +440,7 @@ class TestS3FileStore:
     @mock.patch.object(storage.S3FileStore, "_get_boto_object_for_key")
     def test_get_last_modified(self, get_boto_object_for_key):
         k = mock.Mock()
-        k.last_modified = datetime.datetime.now()
+        k.last_modified = datetime.datetime(2023,1,1,8,0,0,0)
         get_boto_object_for_key.return_value = k
         store = storage.S3FileStore("bucket")
 
--- a/tests/storage/test_storage.py
+++ b/tests/storage/test_storage.py
@@ -1,4 +1,5 @@
 import builtins
+import datetime
 import os
 import shutil
 import tempfile
@@ -436,6 +437,18 @@ class TestS3FileStore:
             OutputSerialization={"CSV": {"FieldDelimiter": ",", "RecordDelimiter": "\n"}},
         )
 
+    @mock.patch.object(storage.S3FileStore, "_get_boto_object_for_key")
+    def test_get_last_modified(self, get_boto_object_for_key):
+        k = mock.Mock()
+        k.last_modified = datetime.datetime.now()
+        get_boto_object_for_key.return_value = k
+        store = storage.S3FileStore("bucket")
+
+        last_modified = store.get_last_modified("a/b/c.pdf")
+        assert get_boto_object_for_key.called
+        assert last_modified == k.last_modified
+        assert type(last_modified) == datetime.datetime
+
     def test_fetch_file_contents_using_s3_select_and_expect_output_in_json_format(self):
         store = storage.S3FileStore("some-bucket")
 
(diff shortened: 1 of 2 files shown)


xocto/storage/storage.py
```diff
@@ -1179,6 +1188,14 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
         file_stats = os.stat(filepath)
         return file_stats.st_size
 
+    def get_last_modified(self, key_path: str) -> datetime.datetime:
+        file_path = os.path.join(self.storage_root, self.bucket_name, key_path)
+        if not os.path.exists(file_path):
+            raise KeyDoesNotExist(f"Key {key_path} was not found at {file_path}")
+
+        file_stats = os.stat(file_path)
+        return datetime.datetime.fromtimestamp(file_stats.st_mtime)
```
durden, line 1197: ah now I get it this is in `LocalFileStore` so this works, sorry missed that before.

xocto/storage/storage.py
```diff
@@ -1440,6 +1440,12 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
         assert s3_object.bucket_name == self.bucket_name
         return len(self.fetch_file_contents(s3_object.key))
 
+    def get_last_modified(self, key_path: str) -> datetime.datetime:
+        """
+        Given this is used for testing purposes, it's not as relevant.  Just return current time.
```
durden, line 1445: Excellent comment!

tests/storage/test_storage.py
```diff
@@ -436,6 +437,18 @@ def test_fetch_file_contents_using_s3_select_and_expect_output_in_csv_format(sel
...
+    def test_get_last_modified(self, get_boto_object_for_key):
+        k = mock.Mock()
+        k.last_modified = datetime.datetime.now()
+        get_boto_object_for_key.return_value = k
+        store = storage.S3FileStore("bucket")
+
+        last_modified = store.get_last_modified("a/b/c.pdf")
+        assert get_boto_object_for_key.called
+        assert last_modified == k.last_modified
+        assert type(last_modified) == datetime.datetime
```
durden, line 450: ❓ if we're verifying it's equal above why do we need to check the type? That feels like a type check situation and what's an example where a datetime (which we set in the mock) compares `==` with something that is _not_ a datetime?

In Python 2 that _could_ work in weird scenarios, but in Python 3 comparing types that don't match results in `TypeError` if memory serves.

tests/storage/test_storage.py
```diff
@@ -436,6 +437,18 @@ def test_fetch_file_contents_using_s3_select_and_expect_output_in_csv_format(sel
             OutputSerialization={"CSV": {"FieldDelimiter": ",", "RecordDelimiter": "\n"}},
         )
 
+    @mock.patch.object(storage.S3FileStore, "_get_boto_object_for_key")
+    def test_get_last_modified(self, get_boto_object_for_key):
+        k = mock.Mock()
+        k.last_modified = datetime.datetime.now()
```
durden, line 443: 🐼 Instead of using `now()` I'd use a concrete date time just to make this test less susceptible to a weird flakey time problem.

xocto/storage/storage.py
```diff
@@ -1423,6 +1440,12 @@ def get_size_in_bytes(self, *, s3_object: S3Object) -> int:
...
+        """
+        return datetime.datetime.now()
```
pydanny, line 1447: Inside a truncated PR-style view this doesn't make any sense, even though it is for memory storage. In a future PR maybe we can clarify things?

pydanny (approved):
LGTM with a small comment about clarity

tests/storage/test_storage.py
```diff
@@ -659,6 +672,16 @@ def test_store_and_fetch_datepath(self):
...
 
+    def test_get_last_modified(self):
+        with tempfile.TemporaryDirectory() as tdir:
+            store = storage.LocalFileStore("bucket", tdir, use_date_in_key_path=True)
+
+            __, path = store.store_file(namespace="x", filename="test.pdf", contents="hello")
+
+            last_modified = store.get_last_modified(path)
+            assert last_modified is not None
+            assert type(last_modified) == datetime.datetime
```
durden, line 683: ♻️ Don't feel like we need to do type checking here.

durden (approved):
Looks good just a few non-blocking suggestions

xocto/storage/storage.py
pydanny: Yup, I had to go digging too, hence my comment.

tests/storage/test_storage.py
pydanny: Oh yes, this is a gotcha I've run into before. This might work on your computer right now, but the next computer might be just fast enough to trigger this as a failure.

tests/storage/test_storage.py
davetran55: Because I want to make sure nothing is funky happens inside the get_last_modified function that screws up with the return type.  Plus I wanted to be explicit in the test that the type I'd like return is a datetime

tests/storage/test_storage.py
davetran55: Same as above about being redundant.

tests/storage/test_storage.py
davetran55: Great suggestion.  Fixed.

xocto/storage/storage.py
davetran55: Thanks.  Will do.

PR #35, opened 2023-01-20
Update version number

aplusclients:
This Pr aims to update version number from 2.2.1 to 2.2.2 due to recent addition of get_last_modified function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 # Changelog
 
-## v2.2.2 - 2023-01-20
+## v2.3.0 - 2023-01-20
 
 Includes:
 
 - Addition of get_last_modified function ([#32](https://github.com/octoenergy/xocto/pull/32))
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -28,7 +28,7 @@
 author = "Kraken Tech"
 
 # The full version, including alpha/beta/rc tags
-release = "2.2.1"
+release = "2.2.2"
```
pydanny, line 31: The addition of new feature isn't a maintenance update. Instead this is an update to `v2.3.0`.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v2.2.2 - 2023-01-20
```
pydanny, line 3: v2.3.0

pydanny (changes requested):
This isn't a maintenance fix, rather a new feature. That justifies bumping to v2.3.0

pydanny (approved):

PR #41, opened 2023-02-27
Add localtime function period_exceeds_one_year

Peter554:
Useful for determining whether two datetimes are more than one year apart. Takes into account leap year and DST issues.

Context: In germany an invoice may not legally cover more than one year of consumption. This functionality is needed in a couple of places within germany logic, and can likely be useful for others too. Motivated by https://github.com/octoenergy/kraken-core/pull/72807 and https://github.com/octoenergy/kraken-core/pull/78884.

### DIFF
--- a/tests/test_localtime.py
+++ b/tests/test_localtime.py
@@ -4,6 +4,7 @@ import decimal
 import pytest
 import pytz
 import time_machine
+from dateutil import relativedelta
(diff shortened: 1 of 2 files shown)


pydanny:
LGTM, in fact I really like how concise and well-documented new code is that has been added.

PR #42, opened 2023-02-28
Bump version to v2.4.0

Peter554:
I followed https://xocto.readthedocs.io/en/latest/xocto/development.html#publishing

`make publish` failed at the `git push` part since `main` is a protected branch...

It was already published to pypi (sorry), since that part came first and did not fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## v2.4.0 - 2023-02-28
+
+Includes:
+
+- Many type annotations added ([#33](https://github.com/octoenergy/xocto/pull/33)).
+- Type annotation for `get_finite_datetime_ranges_from_timestamps` updated ([#37](https://github.com/octoenergy/xocto/pull/37)).
+- Added the `localtime` function `period_exceeds_one_year` for determining whether a datetime period 
+  exceeds one year ([#41](https://github.com/octoenergy/xocto/pull/41)).
+
 ## v2.3.0 - 2023-01-20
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## v2.4.0 - 2023-02-28
+
+Includes:
+
+- Many type annotations added ([#33](https://github.com/octoenergy/xocto/pull/33)).
```
Peter554, line 7: @pydanny @codeinthehole does this count as a breaking change (=> major version)? 

It's certainly causing a bit of friction to upgrade here https://github.com/octoenergy/kraken-core/pull/79614#issuecomment-1448085479. Generally if the client has to modify their code to accept the upgrade I would say it should be a major version. But this is just type annotations, and it's only krakens static analysis / linting which breaks. So, I'm not sure.

The removal/renaming of `StorageFile ` is also technically breaking, I would say https://github.com/octoenergy/kraken-core/pull/79614#issuecomment-1447882987. Although the effort for the client to upgrade is small.

pydanny: In Semver breaking changes are a major version upgrades. I think we should do a 3.0 release.

pydanny (approved):
LGTM, but as @Peter554 brought up breaking changes in semver are a major release, so this should be for 3.0. Since this is up on PyPI already I'll accept this PR. 

We'll do a 3.0 release and I'll see about removing 2.4 from the release list, including PyPI.

PR #46, opened 2023-03-03
Add tracing to xocto

noelia-mari:
Add tracing file which is used to add more trace information to DataDog events.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204105084296395

### DIFF
--- a/xocto/tracing.py
+++ b/xocto/tracing.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+import ddtrace
+
+tracer = ddtrace.tracer
+wrap = tracer.wrap
+
+
+def set_tags(tags: dict[str | bytes, str]) -> None:
+    """
+    Set multiple tags on the current span.
+    """
+    span = tracer.current_span()
+    if span:
+        span.set_tags(tags)
+
+
+def set_tag(key: str, value: object) -> None:
+    """
+    Set a tag on the current span.
+    """
+    span = tracer.current_span()
+    if span:
+        span.set_tag(key, value)
+
+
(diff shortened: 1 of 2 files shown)


xocto/tracing.py
```diff
@@ -0,0 +1,34 @@
...
+
+import ddtrace
+
+tracer = ddtrace.tracer
+wrap = tracer.wrap
+
+
+def set_tags(tags: dict[str | bytes, str]) -> None:
+    """
+    Sets multiple tags on the current span
```
cgl, line 11:
```suggestion
    Set multiple tags on the current span.
```

xocto/tracing.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    span = tracer.current_span()
+    if span:
+        span.set_tags(tags)
+
+
+def set_tag(key: str, value: object) -> None:
+    """
+    Sets a tag on the current span
```
cgl, line 20:
```suggestion
    Set a tag on the current span.
```

xocto/tracing.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    span = tracer.current_span()
+    if span:
+        span.set_tag(key, value)
+
+
+def set_global_tag(key: str, value: object) -> None:
+    """
+    Sets a tag on the current root span. These tags will be associated to the
+    entire Datadog Trace
```
cgl, line 30:
```suggestion
    Set a tag on the current root span. 
    
    These tags will be associated to the entire Datadog trace.
```

cgl (approved):

PR #50, opened 2023-03-07
Bump version to v3.0.1

datur:
Addition of get_git_branch_name function for PACT testing

Many apologies, it looks like this has already been released on PyPi after following the instructions [here](https://xocto.readthedocs.io/en/latest/xocto/development.html#publishing) but committing the bump failed as main is protected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v3.1.0 - 2023-03-07
+
+Includes:
+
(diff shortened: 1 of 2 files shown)


pydanny (changes requested):
In semver addition of a feature is a minor release, not a patch. So this should be 3.1.0.

pydanny:
Closing this as I've done a 3.1.0 release for you. 😄

PR #52, opened 2023-03-13
Add numpy dependency

noelia-mari:
As `storage/files.py` is importing pandas and it needs the numpy dependency xocto should have it as an extra required dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
             "isort==5.11.4",
             "mypy-boto3-s3==1.26.0.post1",
             "mypy==0.991",
+            "numpy==1.22.2",
             "twine==4.0.2",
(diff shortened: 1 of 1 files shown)


pydanny (changes requested):
Thanks for the PR! 🙂

I'm not certain this is necessary and a pinned numpy version might cause conflict downstream. Reading [this](https://github.com/pandas-dev/pandas/blob/main/pyproject.toml) from my phone, Pandas has numpy defined as a dependency across a range of versions.

Of course I might be wrong or you might be looking at something I'm not seeing.

pydanny (approved):
LGTM after talking to @noelia-mari, pandas isn't installing numpy by default.

Separately, I'll raise an issue about having xocto dependencies being a range rather than pinned dependencies.

PR #54, opened 2023-03-20
Accept iterable instead of list

GlennS:
There's no reason this needs to be a `list` in particular.

When [upgrading Kraken's xocto version](https://github.com/octoenergy/kraken-core/pull/79737), this is causing a lot of noise as we need to unnecessarily convert a lot of `set[datetime.datetime]` into `list[datetime.datetime`.


cogat (approved):

GlennS:
I tagged @octoenergy/core-financials since they wrote this originally, but looking again it's a really obvious type change, so I'll merge now and they can always put it back if they disagree.

Actually the first thing we do in the function with this _"list"_ is convert it into a set anyway...

PR #56, opened 2023-03-23
Move ddtrace dependency

cgl:
Prior to this change, ddtrace was in development dependencies, but it is a core dependency for the tracing module.

This change moves ddtrace from dev dependencies to core dependencies.


noelia-mari (approved):
Looks good. 😄  I have a question, should we update the CHANGELOG and bump the version in this same PR? 🤔

PR #57, opened 2023-03-24
Publishing to pypi using GitHub Actions CI/CD workflows

cgl:
Prior to this change, we were manually uploading dist to pypi.

This change introduces a GitHub Actions CI/CD workflow to publish a Python distribution whenever a tagged commit is pushed. It will use the pypa/gh-action-pypi-publish GitHub Action.

https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/


meshy (approved):
:tada:

PR #66, opened 2023-04-25
Add localtime function to translate months to German

laurentputz:
~Copy-pasted from~Inspired by the very similar function `translate_english_month_to_spanish`, this commit adds a small helper to the localtime module that takes in a month as `int` and returns the German month name as `str`.

### DIFF
--- a/xocto/localtime.py
+++ b/xocto/localtime.py
@@ -694,6 +694,25 @@ def translate_english_month_to_spanish(month: int) -> str:
     return month_name_lookup[month_name]
 
 
+def translate_english_month_to_german(month: int) -> str:
+    english_month_name = calendar.month_name[month]
+    month_name_lookup = {
+        "January": "Januar",
+        "February": "Februar",
+        "March": "März",
+        "April": "April",
+        "May": "Mai",
+        "June": "Juni",
+        "July": "Juli",
+        "August": "August",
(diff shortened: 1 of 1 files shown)


xocto/localtime.py
```diff
@@ -694,6 +694,25 @@ def translate_english_month_to_spanish(month: int) -> str:
...
+        "May": "Mai",
+        "June": "Juni",
+        "July": "Juli",
+        "August": "August",
+        "September": "September",
+        "October": "Oktober",
+        "November": "November",
+        "December": "Dezember",
+    }
+    return month_name_lookup[english_month_name]
```
meshy, line 713: Suggestion: We can probably go without the English months entirely, I think:

```suggestion
def translate_month_number_to_german(month: int) -> str:
    month_name_lookup = {
        1: "Januar",
        2: "Februar",
        3: "März",
        4: "April",
        5: "Mai",
        6: "Juni",
        7: "Juli",
        8: "August",
        9: "September",
        10: "Oktober",
        11: "November",
        12: "Dezember",
    }
    return month_name_lookup[month]
```

meshy: Should we handle the case of an integer that's out of range?

laurentputz:
Change not suitable

PR #67, opened 2023-05-01
Add an extra example to `ranges.Range.union`

GlennS:
I expected two ranges like this to `union` into a single range, but they don't!

Documenting that fact with an example.

### DIFF
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -106,7 +106,7 @@ class Range(Generic[T]):
         True
 
     * The `union` function (which is aliased to the or (|) operator), will return a range which
-      covers two overlapping (or adjacent) ranges, or None if they are disjoint.
+      covers two overlapping (or touching) ranges, or None if they are disjoint.
         >>> Range(0, 2).union(Range(1, 3))
         <Range: [0,3)>
         >>> Range(0, 2) | Range(2, 4)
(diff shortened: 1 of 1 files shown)


xocto/ranges.py
```diff
@@ -115,6 +115,8 @@ class Range(Generic[T]):
         None
         >>> Range(0, 2) | Range(2, 4, boundaries="(]")
         None
+        >>> Range(0, 2, boundaries="[]") | Range(3, 4)
+        None
```
cogat, line 119: I'd say this example needs a justification. What is the learning here about using `Range` with integers?

cogat (changes requested):
The point is that `Range` is designed for continuous ranges, so will produce possibly unintuitive results when used on integers, like dates. I think this should be made explicit. Perhaps worth calling out "`Range` will not convert end-inclusive `date` into `datetime` objects."

xocto/ranges.py
GlennS: I agree but I have no idea how to word it.

cogat: Put it at the top, something like

"Note that using `Range` with `date` objects does not convert them to `datetimes`. Thus the end of one date does not overlap with the beginning of the next date like it might in an end-inclusive `datetime` comparison. To achieve this, convert your dates to `datetimes` before creating the range.

GlennS: I'm not sure that's a full description of the issue.

I think it's more something like - the documentation on `union` says that adjacent ranges will be merged, but that's misleading.

So I think the right wording is something to do with a the meaning of _adjacent_ - what it really means is that the ranges have to either start and end on exactly the same value, or overlap a bit. They can't just inclusively touch values which are next to each other.

For example if we had one range which ended (inclusive) 1 millisecond* before midnight, and another which started (inclusive) on exactly midnight, they wouldn't be considered adjacent. They can only be considered adjacent if they're exactly the same value.

*Nanosecond? Whatever the smallest increment is.

GlennS: @cogat I've updated with a 2nd attempt.

cogat (approved):

PR #68, opened 2023-05-02
Allow in memory store copying from different a bucket

GeorgeMattimoe:


tests/storage/test_storage.py
```diff
@@ -621,6 +621,30 @@ def test_download_file(self):
...
+        source_bucket_name, source_key_path = source_store.store_file(
+            namespace="mem",
+            filename="test.pdf",
+            contents=contents,
+        )
+        s3_object = storage.S3Object(
+            bucket_name=source_bucket_name,
+            key=source_key_path,
+        )
+        destination = "a/b/c.pdf"
```
cgl, line 637: 🐼 Since the original key path is called `source_key_path`, shall we rename this as `destination_key_path`?

xocto/storage/storage.py
cgl, line 1: 🐼 There is a nice PR description, shall we copy that as detailed explanatory text  into the commit message? 
Only truly trivial changes should have a one-line commit message, see: [standard commit message format](https://github.com/octoenergy/coding-conventions/blob/0e7d9e725e24f8f4521ac78267f25976ee80a2a7/conventions/pull-requests.md#follow-our-standard-commit-message-format).

cgl (commented):
I think this is going along the right direction. I have a few comments left.

PR #69, opened 2023-05-04
Ranges: add utils to check period covering

Peter554:
Adding ~two~ one utility functions to the ranges module here.

* ~`period_is_fully_covered` checks whether the passed period is fully covered by the passed ranges.~ Already exists in `RangeSet.contains_range`.
* `any_overlapping` checks whether any of the passed ranges overlap.

### DIFF
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -832,3 +832,15 @@ def get_finite_datetime_ranges_from_timestamps(
         # or end of the period.
         if start != end
     ]
+
+
+def any_overlapping(ranges: Sequence[Range[T]]) -> bool:
+    """Return true if any of the passed Ranges are overlapping."""
+    if not ranges:
+        return False
+    range_set = RangeSet[T]([ranges[0]])
+    for range in ranges[1:]:
+        if range_set.intersection(range):
+            return True
+        range_set.add(range)
+    return False
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -561,3 +561,40 @@ class TestBreakPeriodsOnTimestamp:
             ranges.FiniteDatetimeRange(splits[1], splits[0]),
             ranges.FiniteDatetimeRange(splits[0], period.end),
         ]
+
+
+class TestAnyOverlapping:
+    @pytest.mark.parametrize(
+        "ranges_",
+        [
+            [
+                ranges.Range(0, 2),
+                ranges.Range(1, 3),
+            ],
+            [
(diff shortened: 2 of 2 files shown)


xocto/ranges.py
```diff
@@ -832,3 +832,9 @@ def get_finite_datetime_ranges_from_timestamps(
         # or end of the period.
         if start != end
     ]
+
+
+def period_is_fully_covered(period: Range[T], ranges: Sequence[Range[T]]) -> bool:
```
leamingrad, line 837: 🐼 I think the naming here isn't quite right (periods are related to temporal data).

How about `is_subrange`?

leamingrad, line 837: 🤔 Rather than a sequence of ranges, could we make the caller pass in something like `Range[T] | RangeSet[T]` in keeping with the other operations in the module?

xocto/ranges.py
```diff
@@ -832,3 +832,9 @@ def get_finite_datetime_ranges_from_timestamps(
         # or end of the period.
         if start != end
     ]
+
```
leamingrad, line 835: 🤔 I think this function might already exist as [`RangeSet.contains_range`](https://github.com/octoenergy/xocto/blob/main/xocto/ranges.py#L572-L574)?

xocto/ranges.py
```diff
@@ -838,3 +838,14 @@ def period_is_fully_covered(period: Range[T], ranges: Sequence[Range[T]]) -> boo
     range_set = RangeSet[T]([period])
     range_set -= RangeSet[T](ranges)
     return not bool(range_set)
+
+
+def any_overlapping(ranges: Sequence[Range[T]]) -> bool:
```
leamingrad, line 843: 🐼 Could you add a docstring for this function?

leamingrad (changes requested):
These seem like useful additions, but I've added enough small comments that I'd like to re-review.

My concerns are mostly about the API we present here (since its harder to change later if we don't get it right first time).

xocto/ranges.py
Peter554: Yes somehow I missed that this did the same thing, but I checked now and you're right, thanks.

xocto/ranges.py
Peter554: This function was removed -> https://github.com/octoenergy/xocto/pull/69#discussion_r1186008306

Peter554: https://github.com/octoenergy/xocto/pull/69#discussion_r1186008306

xocto/ranges.py
Peter554: Added.

leamingrad (approved):
LGTM 🚢

PR #73, opened 2023-05-23
Consolidate PyPI publishing intructions

codeinthehole:
There were two sets of instructions for publishing to PyPI which are now
consolidated into one. Plus a few other minor improvements along the way.


cgl (approved):
Thanks for the upgraded version, LGTM 🚀

PR #76, opened 2023-06-02
SPA Day: Check type for variable in list of classes

riseriyo:
Prior to this PR, we check if the variable, v, is an instance of the different classes, `FileHeaderInfo, QuoteFields, CompressionType`. It was suggested to use python's `switch` statement which I gave a try, but didn't use. 

This PR uses an `if` conditional to check the type for the variable, v, in the list of above mentioned classes. 

Reference:
[GH link](https://github.com/octoenergy/kraken-core/pull/63061#discussion_r971002498)
[Asana Spa task](https://app.asana.com/0/1200138786534317/1202981523334330)

Images:
<details>
<summary>The catalyst conversation to this Spa Day PR</summary>

Screenshot 2023-06-02 at 9 22 56 AM

</details>

### DIFF
--- a/xocto/storage/s3_select.py
+++ b/xocto/storage/s3_select.py
@@ -58,12 +58,9 @@ class BaseSerializer:
         temp_dict = {k: v for k, v in dataclasses.asdict(self).items() if v}
 
         for k, v in temp_dict.items():
-            if (
-                isinstance(v, FileHeaderInfo)
-                or isinstance(v, QuoteFields)
-                or isinstance(v, CompressionType)
-            ):
+            if type(v) in [FileHeaderInfo, QuoteFields, CompressionType]:
                 temp_dict[k] = v.value
+
         return temp_dict
 
 


xocto/storage/s3_select.py
```diff
@@ -58,12 +58,9 @@ def to_dict(self) -> dict[str, Any]:
         temp_dict = {k: v for k, v in dataclasses.asdict(self).items() if v}
 
         for k, v in temp_dict.items():
-            if (
-                isinstance(v, FileHeaderInfo)
-                or isinstance(v, QuoteFields)
-                or isinstance(v, CompressionType)
-            ):
+            if type(v) in [FileHeaderInfo, QuoteFields, CompressionType]:
```
pydanny, line 61: xocto currently supports Python 3.9, which doesn’t include switch/case - those comes in Python 3.10. So this was the right move. For now. :wink:

pydanny (approved):
LGTM

xocto/storage/s3_select.py
tahakhan0, line 61: I don't think this is the best way because `type` doesn't check for `inheritance`.  This link has a good [answer](https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance)

PR #80, opened 2023-06-08
Remove convenience type aliases

samueljsb:
Prior to this change, these aliases are sometimes used by users to
construct date/datetime objects, with the appearance that they are
constructing a timezone aware object. However, this is not the case,
since they are just aliases for the stdlib:

```pycon
>>> my_datetime = localtime.DateTime(2023, 6, 8, 11, 32)
>>> my_datetime
datetime.datetime(2023, 6, 8, 11, 32)
>>> repr(my_datetime.tzinfo)
'None'
```

This change removes these type aliases. This makes the module more
consistent (the stdlib imports were used for some type annotations) and
removes this footgun.

### DIFF
--- a/xocto/localtime.py
+++ b/xocto/localtime.py
@@ -12,14 +12,6 @@ from django.utils import timezone
 
 from . import numbers, ranges
 
-# Convenience type aliases for referencing from other modules. The 'Union` is redundant and just
-# to help mypy realise that these are type aliases, not assignments.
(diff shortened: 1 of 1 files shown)


meshy (approved):

pydanny (approved):
I was always weirded out by this segment, just couldn't articulate my complaint. Thanks Sam for doing things the right way.

PR #89, opened 2023-09-01
Replace group of isinstance with hasattr

sbj3:
Small Spa day refactor.


xocto/storage/s3_select.py
```diff
@@ -58,11 +58,7 @@ def to_dict(self) -> dict[str, Any]:
         temp_dict = {k: v for k, v in dataclasses.asdict(self).items() if v}
 
         for k, v in temp_dict.items():
-            if (
-                isinstance(v, FileHeaderInfo)
-                or isinstance(v, QuoteFields)
-                or isinstance(v, CompressionType)
-            ):
+            if hasattr(v, "value"):
```
pydanny, line 61: This is taking away validation. Can you provide documentation as a comment on the PR or in the code demonstrating this specification is correct?

pydanny (changes requested):

xocto/storage/s3_select.py
sbj3: Hmm. Had not looked at it as validation. My understanding of this section (for loop) was that it was just for having the display value as opposed to internal value, like "PENDING" instead of the number of the enum. An additional aspect is that these three types are all defined in this repository as sub classes of Enum. :thinking:

pydanny: Thanks for the answer! Without it the purpose of this as a refactor was unclear. 😄

pydanny (approved):
LGTM

PR #90, opened 2023-09-04
Widen type for ranges.any_overlapping

benthorner:
In response to [^1]. Allowing any iterable is more convenient for
the caller of this function.

[^1]: https://github.com/octoenergy/kraken-core/pull/100203#discussion_r1312518191


GlennS (approved):
:+1: thanks @benthorner

PR #91, opened 2023-09-06
Add "parse_date/dt" functions to localtime utility

benthorner:
In response to [^1].

These are especially helpful in testing but also in running code, where we
tend to repeat the lines to parse/make-aware.

[^1]: https://github.com/octoenergy/kraken-core/pull/100203#discussion_r1312523206

### DIFF
--- a/xocto/localtime.py
+++ b/xocto/localtime.py
@@ -729,3 +729,20 @@ def period_exceeds_one_year(start_at: datetime_.datetime, end_at: datetime_.date
         one_year_after_start_at = one_year_after_start_at.replace(month=3, day=1)
 
     return tz_unaware_end_at > one_year_after_start_at
+
+
+def parse_date(value: str) -> datetime_.date:
+    """
+    Returns a datetime.date for a given ISO format date string.
+    """
+    return datetime_.date.fromisoformat(value)
+
+
+def parse_dt(value: str) -> datetime_.datetime:
+    """
+    Returns a datetime.datetime for a given ISO format date/time string.
+
+    The date/time must be naive - not include timezone information.
+    """
+    _datetime = datetime_.datetime.fromisoformat(value)
+    return timezone.make_aware(_datetime)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.2.0 - 2023-09-06
+
+- Add `localtime.parse_date` and `localtime.parse_dt` [#91](https://github.com/octoenergy/xocto/pull/91)
+
 ## v4.1.1 - 2023-09-06
 
 - Allow any iterable type in `ranges.any_overlapping` [#90](https://github.com/octoenergy/xocto/pull/90)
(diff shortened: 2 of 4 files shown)


xocto/localtime.py
```diff
@@ -729,3 +730,13 @@ def period_exceeds_one_year(start_at: datetime_.datetime, end_at: datetime_.date
...
+
+
+def parse_date(value: str) -> datetime_.date:
+    _datetime = date_parser.parse(value, dayfirst=("/" in value))
+    return _datetime.date()
+
+
+def parse_dt(value: str) -> datetime_.datetime:
+    _datetime = date_parser.parse(value, dayfirst=("/" in value))
+    return timezone.make_aware(_datetime)
```
GlennS, line 742: Surprised we're not just using `.from_isoformat` here to avoid ambiguity?

GlennS: Might just ping @leamingrad for his view.

The code looks good though.

benthorner: > Surprised we're not just using .from_isoformat here to avoid ambiguity?

That's tempting as a first pass - I actually didn't realise e.g. `2020-02-02 10:00` is valid ISO (thought it had to have a `T` marker). I don't mind really - would like to keep "parse_dt" naming though as it's shorter.

leamingrad: I agree - I would probably just use `datetime.datetime.from_isoformat`.

The only issue is that it doesn't accept the `Z` timezone in Python 3.10, but you can work around that with string replacement.

benthorner: Fair enough: updated to `from_isoformat` 👍. We can always extend it if we actually want to support more parseable formats, but I've not seen a need for this just yet - and the tests I added still pass ✅.

> The only issue is that it doesn't accept the Z timezone

Shouldn't matter: `parse_dt` expects a naive timestamp anyway.

xocto/localtime.py
```diff
@@ -729,3 +729,20 @@ def period_exceeds_one_year(start_at: datetime_.datetime, end_at: datetime_.date
...
+    return datetime_.date.fromisoformat(value)
+
+
+def parse_dt(value: str) -> datetime_.datetime:
+    """
+    Returns a datetime.datetime for a given ISO format date/time string.
+
+    The date/time must be naive - not include timezone information.
+    """
+    _datetime = datetime_.datetime.fromisoformat(value)
```
leamingrad, line 747: 🐼 Might be worth copying in https://github.com/octoenergy/kraken-core/blob/60a9023bc11b3880e214e0160f3fa433901325a3/src/octoenergy/plugins/territories/aus/meter_readings/gas/supportsite/serializers.py#L152 here, so that this works with all ISO timestamps (the problem is fixed in later versions of python though)

leamingrad (approved):
LGTM!

xocto/localtime.py
benthorner: Shouldn't be necessary:

> The date/time must be naive - not include timezone information.

So we wouldn't want that to work anyway - at least not at the moment. Maybe this should cope with zoned timestamps, but then really you could/should just use `fromisoformat` directly.

PR #93, opened 2023-09-18
Allow timezone override in localtime.parse_dt

benthorner:
Follow up to [^1].

This is necessary for compatibility with "industrytime" wrappers
in kraken-core, which inject the optional "tz" override [^2].

[^1]: https://github.com/octoenergy/xocto/pull/91
[^2]: https://github.com/octoenergy/kraken-core/blob/bf82578340d2cc99b7e7902cd6f7ccd21a7f2f19/src/octoenergy/utils/industrytime.py#L110


GlennS (approved):
:raised_hands: thanks @benthorner

PR #94, opened 2023-09-20
Add instructions on proper versioning

pydanny:
Our changelog doesn't follow semver consistently. This PR adds instructions and references to do so.


delfick (approved):
leaving aside a philosophical discussion about whether a single number can ever be truly be relied on to express a multitude of changes, this is definitely a reasonable addition and good to point out.

jamesbeith (approved):

PR #95, opened 2023-09-25
Enable parquet input for S3FileStore s3 select query

alexeocto:
This PR enables the `S3FileStore.fetch_object_contents_using_s3_select` method to also query parquet files by adding a parquet serializer to the `s3_select.py` file. 

A `LocalFileStore` equivalent method was also implemented in order to facilitate local development, in particular linked to https://github.com/octoenergy/kraken-core/pull/102690 and Integrity checks.

The `LocalFileStore.fetch_object_contents_using_s3_select` works by first loading the entire file from it's local storage location, and then creating a DuckDB in memory.  DuckDB was selected over SQLite and SQLAlchemy as they were unable to handle more complex datatypes such as arrays without peculiarities and hashing whereas DuckDB with all test data returned results in the same format as expected.

Once the temporary DB is created in memory, the file is loaded into a table in the DB, where an SQL query is run against the DB returning a dataframe with the results. This is then output as a CSV or JSON using the arguments from the output serializer class in order to mimic the results that would be returned by s3.

A comprehensive suite of unit tests has then been added.


xocto/storage/storage.py
```diff
@@ -1107,6 +1126,158 @@ def fetch_file(self, key_path: str, version_id: str | None = None) -> StreamingB
...
+
+        yield result
+
+    def query_dataframe_with_sql(
+        self,
+        raw_sql: str,
+        df: pd.DataFrame,
+    ) -> pd.DataFrame:
+        con = duckdb.connect(database=":memory:")
+        con.register("s3_object", df)
```
alexeocto, line 1184: need to change this to s3object

xocto/storage/s3_select.py
```diff
@@ -165,3 +170,25 @@ def get_serializers_for_csv_file(
...
+
+def get_serializers_for_parquet_file(
+    output_serializer: JSONOutputSerializer | CSVOutputSerializer,
+) -> dict[str, dict[str, Any]]:
+    """
+    Returns input and output serialization dictionaries that should be used to perform a select_object_content query.
+
+    Parquet files do not support input serialization or scan ranges.
+    """
+    temp_dict: dict[str, dict[str, Any]] = {
```
jbkkd, line 183: You could just return this directly, without giving it this temp name

```suggestion
    return {
```

xocto/storage/storage.py
jbkkd: Would be better as a comment above this line, with a bit more detail on why you do this

xocto/storage/storage.py
```diff
@@ -1107,6 +1126,158 @@ def fetch_file(self, key_path: str, version_id: str | None = None) -> StreamingB
...
+        record_delimiter = input_serializer.get("RecordDelimiter")
+        header: int | None | str
+
+        if "FileHeaderInfo" in input_serializer.keys():
+
+            if input_serializer["FileHeaderInfo"] == "NONE":
+                header = None
+            elif input_serializer["FileHeaderInfo"] == "IGNORE":
+                raise NotImplementedError(
+                    "This method of column selection not implemented for localdev"
```
jbkkd, line 1215:
```suggestion
                    "Value `IGNORE` of column selection not implemented for localdev"
```

xocto/storage/storage.py
```diff
@@ -1107,6 +1126,158 @@ def fetch_file(self, key_path: str, version_id: str | None = None) -> StreamingB
...
+
+        filtered_df = self.query_dataframe_with_sql(
+            raw_sql=raw_sql,
+            df=df,
+        )
+
+        if isinstance(output_serializer, s3_select.JSONOutputSerializer):
+            if output_serializer.RecordDelimiter != "\n":
+                raise NotImplementedError(
+                    "Record delimiter other than newline not implemented for JSON output in localdev"
```
jbkkd, line 1167:
```suggestion
                    "Only newline ('\n') is supported as the record delimiter for JSON output in localdev"

```

xocto/storage/storage.py
alexeocto: sorry this was just a reminder to me to change this.

jbkkd (approved):

PR #98, opened 2023-10-10
Fix bug in FiniteDateRange.days property

josh-lynch:
Previously this property would return a value exclusive of the end
date. However the FiniteDateRange defines it's boundaries as double
inclusive, therefore this value was off by one.

This change includes the range end date in the count of days.

### DIFF
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -798,7 +798,7 @@ class FiniteDateRange(FiniteRange[datetime.date]):
         """
         Return the number of days between the start and end of the range.
         """
-        return (self.end - self.start).days
+        return (self.end - self.start).days + 1
 
 
 def get_finite_datetime_ranges_from_timestamps(
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -600,3 +600,20 @@ class TestAnyOverlapping:
 
     def test_returns_false_for_empty_set_of_ranges(self):
         assert not ranges.any_overlapping([])
+
+
+class TestFiniteDateRange:
+    """
+    Test class for methods specific to the the FiniteDateRange subclass.
+    """
+
+    def test_days_property_has_correct_value(self):
+        """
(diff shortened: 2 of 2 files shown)


josh-lynch:
This PR addresses a long standing bug in the ranges module. It seems likely that a lot of code will have been developed to depend on this bug, as such we will need to be careful about the release of this change.

For kraken-core (and other in house repos I assume) we can rely on CI picking up any potential breakages this change will cause. However as this repo is now open source I'm not sure how we manage the risk for other users, do we need to consider a major version change?

josh-lynch:
Related PR: https://github.com/octoenergy/xocto/pull/99

xocto/ranges.py
```diff
@@ -798,7 +798,7 @@ def days(self) -> int:
         """
         Return the number of days between the start and end of the range.
         """
-        return (self.end - self.start).days
+        return (self.end - self.start).days + 1
```
meshy, line 801: Thought: This might be one to use the experiments framework on in Kraken. We would be able to observe failures in datadog without causing issues, and correct them as found.

josh-lynch: I'm open to that. That's a problem for step two though (upgrading the xocto version in Kraken). Priority for now is just getting the change accepted here.

meshy: It might be worth adding an old-and-new function here for the transition, so that you can test them side-by-side, otherwise the upgrade will get stuck behind this change.

samueljsb (approved):
I think this is an obvious bug with an obvious fix 👍🏻

xocto/ranges.py
meshy: Thinking about it, I don't think that this is a good candidate for the experiments framework after all, because that's designed around cases where you want both codepaths to have exactly the same result, but we know that we're making a change here. Sorry for the confusion!

leamingrad (approved):
LGTM!

PR #99, opened 2023-10-10
Fix bug in FiniteDateRange.is_disjoint (and union as a result)

josh-lynch:
Primarily this change is aimed at changing the behaviour of the
FiniteDateRange.union. Previously FiniteDateRanges would not make
a union with adjacent ranges, despite their inclusive boundaries
meaning the presence of an adjacent range means they cover a
fully contiguous period of time.

This change makes adjacent FiniteDateRanges unionize by no longer
considering ranges on adjacent days disjoint.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -617,3 +617,175 @@ class TestFiniteDateRange:
             end=datetime.date(2000, 1, 2),
         )
         assert range.days == 2
+
+    class TestIsDisjoint:
+        @pytest.mark.parametrize(
+            "other",
+            [
+                ranges.Range(
+                    start=datetime.date(2000, 1, 1),
+                    end=datetime.date(2000, 1, 2),
+                    boundaries=ranges.RangeBoundaries.EXCLUSIVE_INCLUSIVE,
+                ),
+                ranges.Range(
+                    start=datetime.date(2000, 1, 5),
+                    end=datetime.date(2000, 1, 6),
+                    boundaries=ranges.RangeBoundaries.INCLUSIVE_EXCLUSIVE,
+                ),
+            ],
+        )
+        def test_identifies_ranges_that_are_not_disjoint(self, other):
+            range = ranges.FiniteDateRange(
+                start=datetime.date(2000, 1, 3),
+                end=datetime.date(2000, 1, 4),
+            )
+            assert range.is_disjoint(other) is False
+
+        @pytest.mark.parametrize(
+            "other",
+            [
+                ranges.Range(
+                    start=datetime.date(2000, 1, 1),
+                    end=datetime.date(2000, 1, 2),
+                    boundaries=ranges.RangeBoundaries.INCLUSIVE_EXCLUSIVE,
+                ),
+                ranges.Range(
+                    start=datetime.date(2000, 1, 5),
+                    end=datetime.date(2000, 1, 6),
+                    boundaries=ranges.RangeBoundaries.EXCLUSIVE_INCLUSIVE,
+                ),
+            ],
+        )
+        def test_identifies_ranges_that_are_disjoint(self, other):
+            range = ranges.FiniteDateRange(
+                start=datetime.date(2000, 1, 3),
+                end=datetime.date(2000, 1, 4),
+            )
+            assert range.is_disjoint(other) is True
+
+    class TestUnion:
+        @pytest.mark.parametrize(
+            "range, other",
+            [
+                (
+                    ranges.FiniteDateRange(
+                        start=datetime.date(2000, 1, 1),
+                        end=datetime.date(2000, 1, 2),
+                    ),
+                    ranges.FiniteDateRange(
+                        start=datetime.date(2000, 1, 3),
+                        end=datetime.date(2000, 1, 4),
+                    ),
+                ),
+                (
+                    ranges.FiniteDateRange(
+                        start=datetime.date(2000, 1, 3),
+                        end=datetime.date(2000, 1, 4),
+                    ),
+                    ranges.FiniteDateRange(
+                        start=datetime.date(2000, 1, 1),
+                        end=datetime.date(2000, 1, 2),
+                    ),
+                ),
+            ],
+        )
+        def test_handles_adjacent_ranges(self, range, other):
+            """
+            As FiniteDateRange boundaries are double inclusive, when two ranges
+            end/start on consecutive days they effectively cover a fully contiguous
+            period of time and should make a union.
+            """
+            union = range | other
+            assert union == ranges.FiniteDateRange(
+                start=datetime.date(2000, 1, 1),
+                end=datetime.date(2000, 1, 4),
+            )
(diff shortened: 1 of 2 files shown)


xocto/ranges.py
josh-lynch, line 1: Not quite sure if I should be checking the boundaries on the other range here given the typing doesn't guarantee it will also be double inclusive. 🤔

josh-lynch: I concluded I should and have made the necessary change.

josh-lynch:
I also can't figure out the typing complaints here and they are driving me insane. Insights welcome.

pydanny:
> I also can't figure out the typing complaints here and they are driving me insane. Insights welcome.

What about being more explicit? 

```
    def is_disjoint(self, other: Range[datetime.date]) -> bool:
        # Adjacent dates should not be considered disjoint, we extend the other
        # range to allow them to be considered adjacent.
        other_start = other.start
        if other._is_right_inclusive:
			if other_start is not None:
	            other_start -= datetime.timedelta(days=1)
        other_end = other.end
        if other._is_right_inclusive:
			if other_end is not None:
	            other_end += datetime.timedelta(days=1)
        return super().is_disjoint(
            Range(start=other_start, end=other_end, boundaries=other.boundaries)
        )
```

delfick:
> I also can't figure out the typing complaints here and they are driving me insane. Insights welcome.

```
var2 = var1.blah
assert var1.blah is not None
```
When you do this, mypy still doesn't know if `var2` is not None or not because `var1.blah` might be a different value to `var2`.

so when you do 

```
        other_start = other.start
        if other._is_right_inclusive:
            assert other.start
            other_start -= datetime.timedelta(days=1)
```

what you actually want is

```
        other_start = other.start
        if other._is_right_inclusive:
            assert other_start is not None
            other_start -= datetime.timedelta(days=1)
```

(I have not tested this theory, but I imagine that's what it is)

frederizia:
> what you actually want is
> 
> ```
>         other_start = other.start
>         if other._is_right_inclusive:
>             assert other_start is not None
>             other_start -= datetime.timedelta(days=1)
> ```
> 
> (I have not tested this theory, but I imagine that's what it is)

Yes was just coming here to say this (I have tested it and that sorts it). Other option is doing `other_start = other.start - datetime.timedelta(days=1)` but that's not as nice.

xocto/ranges.py
```diff
@@ -790,6 +790,21 @@ def intersection(self, other: Range[datetime.date]) -> Optional["FiniteDateRange
         assert base_intersection.boundaries == RangeBoundaries.INCLUSIVE_INCLUSIVE
         return FiniteDateRange(base_intersection.start, base_intersection.end)
 
+    def is_disjoint(self, other: Range[datetime.date]) -> bool:
+        # Adjacent dates should not be considered disjoint, we extend the other
+        # range to allow them to be considered adjacent.
+        other_start = other.start
+        if other._is_right_inclusive:
+            assert other.start
+            other_start -= datetime.timedelta(days=1)
```
GlennS, line 799: :cow: this bug is wider than dates - it also affects any `Range` where the values are discrete. For example: integers.

I think it could also potentially affect other boundary conditions than inclusive-inclusive, although I guess we'd have to go through and figure it out.

I wonder if a wider solution would be some kind of `increment` or `resolution` field? Either directly on the `Range` itself, or coming from a mapping `RANGE_TYPE_RESOLUTIONS = {datetime.Date: datetime.timedelta(days=1), int: 1, ...}`.

(I don't want my comment to block this PR. I'm sure date ranges are our biggest problem in Kraken and your fix as it stands would be a very useful win.)

tests/test_ranges.py
```diff
@@ -617,3 +617,135 @@ def test_days_property_has_correct_value(self):
...
+            end/start on consecutive days they effectively cover a fully contiguous
+            period of time and should make a union.
+            """
+            union = range | other
+            assert union == ranges.FiniteDateRange(
+                start=datetime.date(2000, 1, 1),
+                end=datetime.date(2000, 1, 4),
+            )
+
+        def test_handles_ranges_out_of_order(self):
```
GlennS, line 658: :panda_face: I don't follow - isn't this identical to the 2nd case for the test above?

tests/test_ranges.py
```diff
@@ -617,3 +617,135 @@ def test_days_property_has_correct_value(self):
...
+        )
+        def test_handles_disjoint_ranges(self, range, other):
+            """
+            Disjoint ranges (those with a full clear day between them) should not make
+            a union.
+            """
+            union = range | other
+            assert union is None
+
+        @pytest.mark.parametrize(
```
GlennS, line 710: :panda_face: this test is complex enough that it should probably have ids on its cases.

GlennS (approved):
Definitely a :+1: from me - this has been bugging me for a while.

xocto/ranges.py
```diff
@@ -790,6 +790,21 @@ def intersection(self, other: Range[datetime.date]) -> Optional["FiniteDateRange
         assert base_intersection.boundaries == RangeBoundaries.INCLUSIVE_INCLUSIVE
         return FiniteDateRange(base_intersection.start, base_intersection.end)
 
+    def is_disjoint(self, other: Range[datetime.date]) -> bool:
+        # Adjacent dates should not be considered disjoint, we extend the other
+        # range to allow them to be considered adjacent.
+        other_start = other.start
+        if other._is_right_inclusive:
+            assert other.start
```
leamingrad, line 798: ❓ Why does this assertion hold? I would expect `(None, '2022-01-01']` to be a valid range.

Maybe this should be `._is_left_inclusive`?

xocto/ranges.py
leamingrad: I also think this might be a wider problem, and some kind of resolution might be a good idea (although since ranges can be for any orderable type, it might not always work).

I agree that we shouldn't block here though.

leamingrad (changes requested):
This looks sensible to me, but I think it might be bugged in its current form so am requesting changes.

xocto/ranges.py
josh-lynch: I did consider the integer problem here, but I think it's a bit different because if you consider float values then there are plenty of values that exist between 2 and 3.

xocto/ranges.py
josh-lynch: 🤔  You are right, this is a bug, I will review my tests to see why they weren't failing.

tests/test_ranges.py
josh-lynch: 🆗

tests/test_ranges.py
josh-lynch: Yeah I think I meant to delete this in a test refactor, my bad.

josh-lynch:
Ahh @delfick thanks that makes total sense 👌

leamingrad (approved):
LGTM

PR #101, opened 2023-10-12
Replace flake8 and isort with ruff

pydanny:
Bringing xocto into alignment with other Kraken projects.

This identified a typing issue in the `xocto/localtime.py` module, which is why that module has non-formatting changes.


jbkkd (approved):

jbkkd:
Also worth updating the changelog

PR #104, opened 2023-10-17
Upgrade duckdb to 0.9.0

alexeocto:
Newer versions of pandas are incompatible with previous versions of duckdb (I believe up to around 0.8.2), so upgrading duckdb version accordingly to avoid errors.

https://github.com/chroma-core/chroma/issues/1069


josh-lynch:
Change looks fine, but can you split it up into separate commits so we can see what changes have come through in the commit history?

eg.

1. upgrade duckdb
2. bump version

If there's only a bump version commit it means we can't actually see what the changes are from the git log

josh-lynch:
I think you have misunderstood my comment slightly, or possibly somethings just gone a little bit wrong with your git use. The initial commit should literally just be this one line:

```
"duckdb==0.9.0",
```

alexeocto:
Sorry not sure what happened my end there, should be all correctly committed now 🤞

xocto/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "4.5.0"
+__version__ = "4.5.1"
```
pydanny, line 1: 1. Not a PATCH update - what bug does this fix?
2. Don't mix code changes with version updates

pydanny (changes requested):
Unrelated version update in a version update

xocto/__init__.py
alexeocto: If the user had a newer version of `pandas` then using the local filestore version of `select_object_contents` would error. Would that count as a bug?

setup.py
```diff
@@ -35,7 +35,7 @@
     zip_safe=False,
     install_requires=[
         "ddtrace>=1.9.0",
-        "duckdb==0.8.1",
+        "duckdb==0.9.0",
```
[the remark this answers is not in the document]
alexeocto: Noted, have changed that thanks

xocto/__init__.py
alexeocto: I've bumped this to 4.6.0 for now 👍

pydanny (approved):
Let's get this in to help our downstream friends. 😄

PR #106, opened 2023-10-26
Replace ZoneInfo typehints in localtime module

josh-lynch:
Previously we changed the typehints in this module to accept `zoneinfo.ZoneInfo` instances, however this excludes certain objects from being passed that we actually do want to support. This PR changes the typehint to `datetime.tzinfo` objects, which is a more permissive typehint allowing both `ZoneInfo` and the other objects we desire.


jamesbeith (approved):

jamesbeith:
cc. @pydanny

PR #108, opened 2023-10-31
Adding explicit type for Optional Foreign Key

delfick:
In a new enough version of mypy providing an optional to the existing `ForeignKey` type results in an error as `MyModel | None` does not satisfy the `models.Model` that the `Model` TypeVar is bound to


pydanny (approved):
I've been wondering how to make this work. How did you figure this out?

delfick:
> How did you figure this out?
@pydanny I'm a wizard!

PR #111, opened 2023-10-31
Improve autodoc of callables and variables

pydanny:
Currently our docs are useful, but lack visibility in some modules as to what components are provided by the code. This forces the reader to dive into source code. While code diving is a good thing, our docs should better describe what the module does. Types is the worst offender, but Localtime can also be improved.

Also, we can make code diving easier.

What this does:

- Links autodocumented callables with their source code
- Uses the autodata function to document variables in types and localtime


jamesbeith (approved):

delfick (approved):
noice

PR #113, opened 2023-11-01
Provide OptionalOneToOneField

delfick:
In a new enough mypy trying to do `OneToOneField[MyModel | None]` will complain because an optional model does not satisfy the bound condition of the TypeVar we are using here.

This is the OneToOneField equivalent of https://github.com/octoenergy/xocto/pull/108, which I was meant to add but forgot.


s-cooper18 (approved):
LGTM

PR #116, opened 2023-11-01
Modernize setup to use pyproject.toml

pydanny:
- Ensures we match dependencies between environments (main, dev, and docs)
- Simplifies the configuration
- Makes it MUCH easier to break this up into multiple packages

Note: Even though the project in theory had been converted to ruff for linting and formatting, this PR uncovered that some modules were formatted with it and others were not. Hence the volume of files changed.

# Why main, dev, and docs for dependencies?

- "Main" is the `project.dependencies` setting and is the PyPI deployed project expects
- Development and tests require the same dependencies, might as well combine them
- Docs are isolated from development and tests in order to speed up docs generation on Read the Docs. When checking if docs are building or failing, it's really nice to have a fast configuration.


tests/storage/test_files.py
```diff
@@ -24,16 +24,17 @@ def test_returns_size(self, file, size):
 def test_files_hash_correctly():
     file1 = io.BytesIO(b"This is my first file")
     assert (
-        files.hashfile(file1) == "2f1b1b913ca382ad8f992ec6a18ecedfa2fcd8ff21b0a2227614a7bd94c23d2d"
```
codeinthehole, line 27: Strange commit message. Does this mean "make code conform to `ruff check`"? or something like that?

codeinthehole: Also, as per our atomic commits convention - this commit should be squashed into the one that introduces these linting errors.

codeinthehole (changes requested):
Couple of small comments.

There's still a few references to `setup.py` in the codebase to factor out too. 

Would be good to see this repo harmonise with https://github.com/octoenergy/cookiecutter-kraken-package - we could backport some of the docs set-up to there.

pyproject.toml
```diff
@@ -1,5 +1,84 @@
...
+
+[project.urls]
+changelog = "https://github.com/octoenergy/xocto/blob/main/CHANGELOG.md"
+documentation = "https://xocto.readthedocs.io"
+issues = "https://github.com/octoenergy/xocto/issues"
+
+[tool.setuptools]
+packages = ["xocto", "xocto.events", "xocto.storage"]
+# Old setup.py version required
+#  package_data={"xocto": ["py.typed"]},
```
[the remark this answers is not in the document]
pydanny: I'll remove it shortly. Using it as a reference for confirming everything is working as expected.

tests/storage/test_files.py
pydanny: Squashed!

.pre-commit-config.yaml
```diff
@@ -16,22 +16,6 @@ repos:
         stages: [commit]
       - id: trailing-whitespace      # Trims trailing whitespace.
 
-
-  - repo: https://github.com/timothycrosley/isort
```
pydanny, line 20: Thanks for the catch on pre-commit. My VSCode doesn't search in this file. Need to fix that.

.circleci/config.yml
```diff
@@ -7,28 +7,22 @@ jobs:
     steps:
```
pydanny, line 7: Looking at this I like the bold changes. I think I get too timid with these kinds of chores, trying to keep them tiny in scope.

pyproject.toml
```diff
@@ -1,5 +1,83 @@
...
+  {name = "Kraken Technologies", email = "talent@octopus.energy"},
+]
+classifiers = [
+  "Development Status :: 5 - Production/Stable",
+  "Intended Audience :: Developers",
+  "Topic :: Software Development :: Build Tools",
+  "License :: OSI Approved :: MIT License",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
```
jamesbeith, line 24: We should add support for Python 3.12 at some point too.

jamesbeith (approved):

pyproject.toml
pydanny: Agreed, created #121.

Issue #122, opened 2023-11-16
RangeSet does not work correctly for more than two FiniteDateRanges


jarshwah:
Using RangeSet to merge two adjacent FiniteDateRanges works as you'd expect, but attempting to merge three adjacent FiniteDateRanges does not.

DateRanges are discrete not continuous, and are typically used as `[inclusive, inclusive]`. When we have two date ranges that are adjacent, they cover *every date*, and should form a single range, which is indeed what happens. The same should also apply for 3 date ranges, but as demonstrated below, it does not.

I'm working on putting together a PR that demonstrates the issue with tests which are essentially copied below:

```
import datetime
from xocto import ranges
date = datetime.date

def test_two_ranges_combine():
        r1 = ranges.FiniteDateRange(date(2023, 1, 1), date(2023, 1, 10))
        r2 = ranges.FiniteDateRange(date(2023, 1, 11), date(2023, 1, 20))
        merged = ranges.RangeSet([r1, r2])
        assert ranges.FiniteDateRange(date(2023, 1, 1), date(2023, 1, 20)) in merged


 def test_three_ranges_combine():
        # This test fails, because RangeSet is unable to correctly merge more than 2 ranges.
        r1 = ranges.FiniteDateRange(date(2023, 1, 1), date(2023, 1, 10))
        r2 = ranges.FiniteDateRange(date(2023, 1, 11), date(2023, 1, 20))
        r3 = ranges.FiniteDateRange(date(2023, 1, 21), date(2023, 1, 31))
        merged = ranges.RangeSet([r1, r2, r3])
        assert ranges.FiniteDateRange(date(2023, 1, 1), date(2023, 1, 31)) in merged

```

PR #123, opened 2023-11-16
Enable more than two adjacent FiniteDateRanges to be unioned

jarshwah:
Prior to this change, FiniteDateRange was able to correctly union with an adjacent FiniteDateRange, however, the union produced was of the base range type and was then unable to union with a successive adjacent range.

This change ensures that unioning two FiniteDateRanges produces a FiniteDateRange so that it can be unioned with a successive adjacent ranges.

Fixes https://github.com/octoenergy/xocto/issues/122

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -757,6 +757,56 @@ class TestFiniteDateRange:
                 end=datetime.date(2000, 1, 4),
             )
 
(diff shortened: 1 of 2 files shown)


jamesbeith (approved):

josh-lynch:
👍

PR #127, opened 2023-11-20
Package py.typed files in the package data

jarshwah:
Setup tools doesn't automatically include py.typed files in the package
data until the next release after Nov 19, 2023, which means that types
that were previously exported now present as Any.
    
This change adds py.typed files to each of the sub-packages, and includes
them in the package data for each.

https://setuptools.pypa.io/en/latest/history.html#draft-unreleased-as-on-today

https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,11 @@ issues = "https://github.com/octoenergy/xocto/issues"
 [tool.setuptools]
 packages = ["xocto", "xocto.events", "xocto.storage"]
 
+[tool.setuptools.package-data]
+"xocto" = ["py.typed"]
+"xocto.events" = ["py.typed"]
+"xocto.storage" = ["py.typed"]
+
 
 # Mypy
 # ----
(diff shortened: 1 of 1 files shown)


jamesbeith (approved):

pyproject.toml
```diff
@@ -75,6 +75,11 @@ issues = "https://github.com/octoenergy/xocto/issues"
 [tool.setuptools]
 packages = ["xocto", "xocto.events", "xocto.storage"]
 
+[tool.setuptools.package-data]
+"xocto" = ["py.typed"]
+"xocto.events" = ["py.typed"]
+"xocto.storage" = ["py.typed"]
```
jarshwah, line 81: TBH I'm not *certain* that we need these files for each of the packages listed. Mypy isn't clear on this.

codeinthehole: Would `[xocto] = ["**/py.typed"]` be more future-proof?

PR #132, opened 2024-01-03
Add utility fnctions to convert date/time ranges

benthorner:
Needed for https://github.com/octoenergy/kraken-core/pull/116552

This adds two new functions:

- as_finite_datetime_periods
- as_date_periods

Together they can be used to convert (finite) date/time periods to finite
date periods, which is useful where source data is available at time
granularity but needs to be used in date-oriented code.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -898,3 +898,54 @@ class TestFiniteDateRange:
             )
             union = range | other
             assert union == range
+
+
+class TestAsFiniteDatetimePeriods:
+    def test_converts(self):
+        actually_finite_range = ranges.DatetimeRange(
+            datetime.datetime(2020, 1, 1),
+            datetime.datetime(2020, 1, 5),
+        )
+
+        results = list(
+            ranges.as_finite_datetime_periods(
+                [actually_finite_range],
+            )
+        )
+
+        assert results == [
+            ranges.FiniteDatetimeRange(
+                actually_finite_range.start,
+                actually_finite_range.end,
+            )
+        ]
+
+    def test_errors_if_infinite(self):
+        with pytest.raises(ValueError) as exc_info:
+            ranges.as_finite_datetime_periods(
+                [
+                    ranges.DatetimeRange(datetime.datetime(2020, 1, 1), None),
+                ],
+            )
+
+        assert "Period is not finite at start or end or both" in str(exc_info.value)
+
+
+class TestAsDatePeriods:
+    def test_converts(self):
+        dt_range = ranges.FiniteDatetimeRange(
+            datetime.datetime(2020, 1, 1),
+            datetime.datetime(2020, 1, 5),
+        )
+
+        results = ranges.as_date_periods(
+            [dt_range],
+            date_converter=lambda dt: dt.date(),
+        )
+
+        assert results == [
+            ranges.FiniteDateRange(
+                datetime.date(2020, 1, 1),
+                datetime.date(2020, 1, 4),
+            )
+        ]
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -440,10 +440,6 @@ class Range(Generic[T]):
         return self.is_left_finite() and self.is_right_finite()
 
 
-# Type aliases for common range types
-DatetimeRange = Range[datetime.datetime]
-
-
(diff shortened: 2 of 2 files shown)


tests/test_ranges.py
```diff
@@ -898,3 +898,54 @@ def test_doesnt_extend_union(self):
...
+
+        results = ranges.as_date_periods(
+            [dt_range],
+            date_converter=lambda dt: dt.date(),
+        )
+
+        assert results == [
+            ranges.FiniteDateRange(
+                datetime.date(2020, 1, 1),
+                datetime.date(2020, 1, 4),
```
benthorner, line 949: 🧀 Date ranges are inclusive at their ends, so subtract a day.

xocto/ranges.py
```diff
@@ -945,3 +946,56 @@ def any_overlapping(ranges: Iterable[Range[T]]) -> bool:
             return True
         range_set.add(range)
     return False
+
+
+def as_finite_datetime_periods(
+    periods: Iterable[HalfFiniteDatetimeRange | DatetimeRange],
```
benthorner, line 952: 🧀 One source of multiple periods is where we're working with `RangeSet`s, which are essentially lists of periods.

xocto/ranges.py
```diff
@@ -944,3 +946,56 @@ def any_overlapping(ranges: Iterable[Range[T]]) -> bool:
...
+    *,
+    date_converter: Callable[[datetime.datetime], datetime.date],
+) -> Sequence[FiniteDateRange]:
+    """
+    Converts the given date/time periods to date periods.
+
+    The given date converter is used to decide what date to ascribe to a
+    given date/time, since we lose timezone info in this conversion.
+
+    We need to decide what to do with partial days at the start and end:
```
benthorner, line 985: 🧀 These are heuristics that are (I think) commonly what we want, but may still be surprising - another approach would be to (optionally) error if we have partial days.

xocto/ranges.py
```diff
@@ -945,3 +946,56 @@ def any_overlapping(ranges: Iterable[Range[T]]) -> bool:
...
+def as_date_periods(
+    periods: Iterable[FiniteDatetimeRange],
+    *,
+    date_converter: Callable[[datetime.datetime], datetime.date],
+) -> Sequence[FiniteDateRange]:
+    """
+    Converts the given date/time periods to date periods.
+
+    The given date converter is used to decide what date to ascribe to a
+    given date/time, since we lose timezone info in this conversion.
```
leamingrad, line 983: 🤔 I don't quite understand this part, could you clarify the following:

My understanding is that we should be passing around datetimes which have timezone information everywhere, in which case we can use `localtime.date` to get the correct day as needed.

I guess the issue is that the typing doesn't prevent us from passing in an unzoned datetime, but maybe we could just throw a value error in that case?

xocto/ranges.py
leamingrad: I think the current behaviour is a bit surprising, since we treat the start and end differently (e.g. if we go from midday to midday, we keep the first day and drop the last).

To keep things simple, I'd be happy to error if we aren't midnight aligned.

Alternatively, I think we could have two functions: `as_covering_date_periods` and `as_covered_date_periods`, which more explicitly cover both behaviours?

leamingrad (changes requested):
This looks good, but I have some questions about the date conversion function that I would like to bottom out.

If it helps, I'd be happy to quick approve a PR with the first commit and the `as_finite_datetime_periods` function?

benthorner:
Thanks for the review @leamingrad ⭐.

> I'd be happy to quick approve a PR with the first commit and the as_finite_datetime_periods function?

Good idea. I've made a PR with just that: https://github.com/octoenergy/xocto/pull/134

(There's also https://github.com/octoenergy/xocto/pull/133 for a little test reorg that meant this PR was more effort than it needed to be.)

> I think we could have two functions: as_covering_date_periods

Nice name suggestion ➕. However, we still need a date converter:

> we can use localtime.date to get the correct day as needed

For my purpose I need `industrytime.date` instead. On reflection that function is probably small and specific enough to just remain in the domain area I'm working on. What do you reckon, still worth adding here?

PR #133, opened 2024-01-03
Encapsulate existing range tests in classes

benthorner:
Tangent from https://github.com/octoenergy/xocto/pull/132

This makes it clearer what is being tested and helps guide future
developers to add new tests in the same location.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -41,272 +41,265 @@ def valid_integer_range(draw):
     return ranges.Range(start, end, boundaries=boundaries)
 
 
-def test_creation():
-    assert ranges.Range(
-        0, 2, boundaries=ranges.RangeBoundaries.EXCLUSIVE_INCLUSIVE
-    ) == ranges.Range(0, 2, boundaries="(]")
+class TestRange:
+    def test_creation(self):
+        assert ranges.Range(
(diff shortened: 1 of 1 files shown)


tests/test_ranges.py
```diff
@@ -41,272 +41,265 @@ def valid_integer_range(draw):
...
-    ],
-)
-def test_range_union(a_str, b_str, expected_str):
-    a = _range_from_string(a_str)
-    b = _range_from_string(b_str)
-    if expected_str is not None:
-        expected = _range_from_string(expected_str)
-    else:
-        expected = None
+class TestRange:
```
benthorner, line 44: 🧀 This is a awkward diff but the only change I'm making is just:

- Creating a new `TestX` class for the class of functions under test.
- Indenting the existing test code to be within these classes.
- Adding a `self` parameter to each encapsulated test function.

The only existing to this is `test_finite_range` (see below).

tests/test_ranges.py
```diff
@@ -898,3 +895,9 @@ def test_doesnt_extend_union(self):
             )
             union = range | other
             assert union == range
+
+    class TestIn:
```
benthorner, line 850: 🧀 Here we already had a grouping for the class of functions under test, with further classes to encapsulate each function; so here I'm following that pattern for a this test, which was previously at the top-level.

leamingrad (approved):
LGTM!

PR #134, opened 2024-01-11
Add "as_finite_datetime_periods" fn

benthorner:
Needed for octoenergy/kraken-core#116552

This can be used as part of a broader conversion of (finite)
date / time periods to finite date periods - the date part of
the conversion is too simple / specific to add here as well.


leamingrad (approved):
LGTM!

PR #136, opened 2024-02-24
Add new date and datetime based range fields that work with xocto.range types

jarshwah:
The range fields provided by `django.contrib.postgres.fields.ranges` depend on types within `psycopg2.extras` that aren't all that intuitive and shouldn't be exposed too heavily within a regular application.

The range types within `xocto.ranges` are more ergonomic and are designed to be used throughout an application.

This change adapts the postgres contrib range fields to work nicely with xocto.ranges.

Fields implemented in this change:

- FiniteDateRangeField
- FiniteDateTimeRangeField
- HalfFiniteDateTimeRangeField

Notably missing is: `HalfFiniteDateRangeField` because we don't have a corresponding type in xocto.ranges. We already have usage of a half-open daterange field within our main application, so it would be a useful addition later.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,6 @@ repos:
         args: [ --fix ]
       # Run the formatter.
       - id: ruff-format
-        args: [ --fix ]
 
   - repo: local
     hooks:
--- a/docs/xocto/model_fields.md
+++ b/docs/xocto/model_fields.md
@@ -92,6 +92,9 @@ Type: [xocto.ranges.FiniteDatetimeRange](xocto.ranges.FiniteDatetimeRange)
 A field that represents an inclusive-exclusive `[)` ranges of timezone-aware
 datetimes. Both the start and end of the range must not be `None`.
 
+The values returned from the database will always be converted to the local timezone
+as per the `TIME_ZONE` setting in `settings.py`.
+
 ```python
 import datetime
 from django.db import models
@@ -128,6 +131,9 @@ Type: [xocto.ranges.HalfFiniteDatetimeRange](xocto.ranges.HalfFiniteRange)
 
 A field that represents an inclusive-exclusive `[)` ranges of timezone-aware
 datetimes. The end of the range may be open-ended, represented by `None`.
+
+The values returned from the database will always be converted to the local timezone
+as per the `TIME_ZONE` setting in `settings.py`.
 
 ```python
 import datetime
--- a/docs/xocto/model_fields.md
+++ b/docs/xocto/model_fields.md
@@ -92,3 +92,40 @@ assert calendar_entry.event_time == ranges.FiniteDatetimeRange(
 )
 assert calendar_entry.event_time.start == localtime.datetime(2020, 1, 1, 14, 30)
 ```
+
+
+#### HalfFiniteDateTimeRangeField
+
+Module: `xocto.fields.postgres.ranges.HalfFiniteDateTimeRangeField`\
+Bounds: `[)`\
+Type: [xocto.ranges.HalfFiniteDatetimeRange](xocto.ranges.HalfFiniteRange)
+
+> **_NOTE:_** docs can not link directly to `HalfFiniteDatetimeRange` at this stage as it's a type alias
+
+A field that represents an inclusive-exclusive `[)` ranges of timezone-aware
+datetimes. The end of the range may be open-ended, represented by `None`.
+
+```python
+import datetime
+from django.db import models
+from xocto import ranges, localtime
+from xocto.fields.postgres import ranges as db_ranges
+
+class Agreement(models.Model):
+    ...
+    period = db_ranges.HalfFiniteDateTimeRangeField()
+
+agreement = Agreement.objects.create(
+    period=ranges.HalfFiniteDatetimeRange(
+        start=localtime.datetime(2020, 1, 1, 14, 30),
+        end=None,
+    ),
+    ...
+)
+
+assert agreement.period == ranges.HalfFiniteDatetimeRange(
+    start=localtime.datetime(2020, 1, 1, 14, 30),
+    end=None
+)
+assert agreement.period.start == localtime.datetime(2020, 1, 1, 14, 30)
+```
--- a/docs/xocto/model_fields.md
+++ b/docs/xocto/model_fields.md
@@ -24,6 +24,30 @@ not require any changes to the database schema. The migration file generated wil
 to the new fields, but since the underlying database type is the same, the migration will
 be a no-op.
 
+The standard Django query operators are almost the same as for the built-in types.
+They accept `xocto.ranges` as arguments, but don't support passing in a tuple of values:
+
+```python
+
+assert SalesPeriod.objects.filter(
+    period__contains=ranges.FiniteDateRange(
+        start=datetime.date(2020, 1, 10),
+        end=datetime.date(2020, 1, 20),
+    )
+).exists()
+
+assert SalesPeriod.objects.filter(
+    period__overlaps=ranges.FiniteDateRange(
+        start=datetime.date(2020, 1, 10),
+        end=datetime.date(2020, 1, 20),
+    )
+).exists()
+
+# ERROR! This will raise a TypeError
+SalesPeriod.objects.filter(period__overlaps=(datetime.date(2020, 1, 10), datetime.date(2020, 1, 20)))
+
+```
+
 #### FiniteDateRangeField
 
 Module: `xocto.fields.postgres.ranges.FiniteDateRangeField`\
--- a/docs/xocto/model_fields.md
+++ b/docs/xocto/model_fields.md
@@ -1,0 +1,54 @@
+# Model Fields
+
+Custom [Django model fields](https://docs.djangoproject.com/en/dev/ref/models/fields/).
+
+
+## Postgres specific fields
+
+Fields that can only be used with a Postgres database, enhancing the functionality
+provided by [django.contrib.postgres.fields](https://docs.djangoproject.com/en/dev/ref/contrib/postgres/fields/).
+
+### Ranges
+
+Provides [range fields](https://docs.djangoproject.com/en/dev/ref/contrib/postgres/fields/#range-fields) that work
+with the `Range` classes from [xocto.ranges](./ranges.md).
+
+The underlying Range type exposed by the Django ORM is `psycopg2.extras.Range` which is awkward to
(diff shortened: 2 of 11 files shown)


xocto/fields/ranges.py
```diff
@@ -0,0 +1,68 @@
...
+from django.db import models
+
+from xocto import ranges
+
+
+class FiniteDateRangeField(pg_fields.DateRangeField):
+    """
+    A DateRangeField with Inclusive-Inclusive [] bounds.
+
+    The underlying postgres type will always store the range as [). This field
```
delfick, line 20: I wonder if this comment should expand on the notation you've used here

xocto/fields/ranges.py
```diff
@@ -0,0 +1,68 @@
...
+    def from_db_value(
+        self,
+        value: Optional[pg_ranges.DateRange],
+        expression: object,
+        connection: object,
+    ) -> Optional[ranges.FiniteDateRange]:
+        if value is None:
+            return None
+        # DateRange is always stored as [) in the database, so we need to remove
+        # a day from the upper bound to make it inclusive.
```
delfick, line 42: I don't fully understand whether that would have undesirable consequences or not, is there more that can be added to this comment to explain the choice of a day?

xocto/fields/ranges.py
```diff
@@ -0,0 +1,68 @@
...
+            end=self.base_field.to_python(obj["end"]),
+        )
+
+    def value_to_string(self, obj: models.Model) -> Optional[str]:
+        value: Optional[ranges.FiniteDateRange] = self.value_from_object(obj)
+        if value is None:
+            return None
+        base_field = self.base_field
+        start = pg_utils.AttributeSetter(base_field.attname, value.start)
+        end = pg_utils.AttributeSetter(base_field.attname, value.end)
```
delfick, line 62: I'm unsure what these two lines are doing

pyproject.toml
```diff
@@ -49,6 +49,7 @@ dev = [
   "mypy==0.991",
   "numpy==1.22.2",
   "pre-commit>=3.2.0",
+  "psycopg2>=2.8.4",
```
delfick, line 52: given it's an open source library, should we make the postgres stuff optional? and then also make the CI run against both postgres and sqlite?

delfick (commented):
I think this would be a great addition :)

xocto/fields/ranges.py
jarshwah: They are partially cargo-culted from the underlying DateRange type. Initially I left them out, but `base_field` (DateTimeField) calls `value_from_object` which essentially does `getattr(self, attname)` (yay descriptors). This sets up the correct attributes so that the `getattr` works.

pyproject.toml
jarshwah: This is dependency *is* optional. I don't think there's any real value in running tests against both PG and sqlite. We don't have *that* much in xocto that depends on the database, but these fields absolutely depend on postgres. It'll be noted in the docs (once written).

Is there value in namespacing the `fields` folder into a `fields/postgres` structure so folks don't unintentionally import something they can't use?

xocto/fields/ranges.py
jarshwah: I'm not keen to be repeating Word-Word [symbol-symbol) everywhere. I'm thinking of defaulting to just symbols as that's common-enough symbolism. Thoughts?

xocto/fields/ranges.py
jarshwah: I can expand on the comment, yes. But, for your understanding prior to that (and using numbers for terseness):

If we have `[1, 10]`, postgres will store that as `[1, 11)` AND return it as `[1, 11)`. We deduct 1 to get back to [1, 10]`. This is how all discrete ranges work in postgres https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE

pyproject.toml
```diff
@@ -66,6 +66,7 @@ dev = [
 docs = [
   "Sphinx==4.5.0",
   "myst-parser==0.18.1",
+  "sphinx-rtd-theme==2.0.0",
```
jarshwah, line 69: This is required to build the docs locally.

.pre-commit-config.yaml
```diff
@@ -34,7 +34,6 @@ repos:
         args: [ --fix ]
       # Run the formatter.
       - id: ruff-format
-        args: [ --fix ]
```
jarshwah, line 37: This was incorrectly added in a previous change but didn't complain for some reason.

jarshwah (commented):
🍷 🧀 👨‍🍳

pyproject.toml
```diff
@@ -65,6 +66,8 @@ dev = [
 docs = [
   "Sphinx==4.5.0",
   "myst-parser==0.18.1",
+  "sphinxcontrib-serializinghtml==1.1.5",
```
jarshwah, line 69: Extracting these changes into https://github.com/octoenergy/xocto/pull/139

xocto/fields/ranges.py
delfick: Perhaps a full explanation somewhere and a reference to that where you use them?

xocto/fields/ranges.py
delfick: and the range is always in full days?

pyproject.toml
delfick: ah it is too, sorry, didn't see which list it was in.

The namespacing sounds like a reasonable idea

> don't think there's any real value in running tests against both PG and sqlite.
mmmkay

xocto/fields/ranges.py
jarshwah: DateRange is, yes 👍 

DateTimeRanges are continuous and used with `[)` bounds, so there's no need for canonicalisation

tests/fields/postgres/test_ranges.py
```diff
@@ -163,3 +163,113 @@ def test_timezone_conversions(self):
...
+        Timezones are converted correctly when round tripping.
+        """
+        half_finite_datetime_range_melb = ranges.HalfFiniteDatetimeRange(
+            start=datetime.datetime(2024, 1, 10, tzinfo=TZ_MELB),
+            end=None,
+        )
+        obj = models.HalfFiniteDateTimeRangeModel.objects.create(
+            half_finite_datetime_range=half_finite_datetime_range_melb,
+        )
+        half_finite_datetime_range_london = ranges.HalfFiniteDatetimeRange(
```
delfick, line 348: should the test confirm that timezone.localtime doesn't use melbourne itself?

tests/fields/postgres/test_ranges.py
```diff
@@ -163,3 +163,113 @@ def test_timezone_conversions(self):
...
+            start=timezone.localtime(half_finite_datetime_range_melb.start),
+            end=None,
+        )
+        obj.refresh_from_db()
+        assert (
+            obj.half_finite_datetime_range
+            == half_finite_datetime_range_london
+            == half_finite_datetime_range_melb
+        )
+        assert obj.half_finite_datetime_range.start.tzinfo != TZ_MELB
```
delfick, line 358: related to the other comment on this test, this doesn't confirm that it's the local timezone that it converted to, only that it's not melbourne...

delfick (approved):
Would recommend more than just one approve, but I like this

jarshwah:
Thanks @delfick - I've tagged a few folks that have implemented range fields recently to have a look

xocto/fields/postgres/ranges.py
```diff
@@ -0,0 +1,220 @@
...
+        end = pg_utils.AttributeSetter(base_field.attname, value.end)
+        return json.dumps(
+            {
+                "start": base_field.value_to_string(start),
+                "end": base_field.value_to_string(end),
+            }
+        )
+
+
+class HalfFiniteDateTimeRangeField(pg_fields.DateTimeRangeField):
```
a-musing-moose, line 152: **question:** ‏Why no `HalfFiniteDateRangeField`

You the the `FiniteDateRange` to represent an inclusive-inclusive `Date` range with a mandatory start and end. Why not also have an inclusive-exclusive `Date` range with a potentially open end?

docs/xocto/model_fields.md
```diff
@@ -0,0 +1,161 @@
...
+
+```python
+import datetime
+from django.db import models
+from xocto import ranges
+from xocto.fields.postgres import ranges as db_ranges
+
+class SalesPeriod(models.Model):
+    ...
+    period = db_ranges.FiniteDateRangeField()
```
a-musing-moose, line 68: **suggestion:** Perhaps an example or 2 of using these range fields in constraints

One of the major benefits of Posgres ranges are that they make it a lot easier to implement non-overlapping constraints. Some examples of how to use these range fields for that would be handy. ‏

tests/fields/postgres/test_ranges.py
```diff
@@ -0,0 +1,364 @@
...
+                    datetime.date(2024, 1, 15),
+                )
+            )
+        with pytest.raises(TypeError):
+            models.FiniteDateRangeModel.objects.filter(
+                finite_date_range__contains=(
+                    datetime.date(2024, 1, 11),
+                    datetime.date(2024, 1, 15),
+                )
+            )
```
a-musing-moose, line 82: **nitpick (non-blocking):** This is 3 tests hidden in a trench-coat

There are 3 separate tests here - one for each of the filter lookups. Whilst they are all effectively the same check under the hood, tests should ideally only test 1 thing to ensure that they are specific and can more easily pinpoint an issue to the specific area effected.

a-musing-moose (approved):
A couple of questions/suggestions but nothing blocking. This is going to be handy. Thanks

xocto/fields/postgres/ranges.py
jarshwah: Good question, but already addressed in the PR description:

> Notably missing is: HalfFiniteDateRangeField because we don't have a corresponding type in xocto.ranges. We already have usage of a half-open daterange field within our main application, so it would be a useful addition later.

docs/xocto/model_fields.md
jarshwah: I plan to add a set of useful constraints in a follow up PR, and will document them properly once done

tests/fields/postgres/test_ranges.py
jarshwah: Fair, but we don't expect these to ever fail, as they all go via the same codepath. It's more a demonstration of that.

jmurphyau:
🤔

Issue #85, opened 2023-07-11
Raise exceptions instead of using assertions


MocktaiLEngineer:
https://github.com/octoenergy/xocto/blob/e0446deeb17e333a95d1c27cbf268698efbde3f6/xocto/numbers.py#L21

Assertions can't be relied upon for control flow because they can be disabled, as per the following - 

```
$ python --help
usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
...
-O     : remove assert and __debug__-dependent statements; add .opt-1 before
         .pyc extension; also PYTHONOPTIMIZE=x
-OO    : do -O changes and also discard docstrings; add .opt-2 before
         .pyc extension
...
```

Would you be open to me changing these assertions to other appropriate exceptions as I encounter them?

thejasmeetsingh:
Hey @samueljsb 
Just raised a PR with regards to this issue: #137

PR #137, opened 2024-02-28
Raised meaningful exceptions instead of assert statements

thejasmeetsingh:
The changes included in this PR are done with regards to this raised issue #85

### DIFF
--- a/xocto/numbers.py
+++ b/xocto/numbers.py
@@ -21,7 +21,7 @@ def quantise(
     0
     """
     if base <= 0:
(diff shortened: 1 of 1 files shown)


samueljsb (commented):
Thanks for this contribution! It looks like you're our first contributor from a fork, so I think we need to fix some setting to make sure the tests run on this branch -- sorry for the delay while we sort that out.

Would you mind adding tests for this new behaviour, please? We'd also like to keep the commit history a bit cleaner, so if you could squash the commits into one that would be appreciated.

PR #141, opened 2024-03-01
Improve sphinx configuration

pydanny:
This PR does the following:

1. Update Sphinx package versions so docs actually build locally and on ReadTheDocs
2. Update copyrights upon build
3. Updates the xocto version in Sphinx upon build

Thanks to the indomitable @jarshwah for flagging this needed to be done. :power:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,9 +11,15 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
 import os
+import pathlib
 import sys
+from datetime import datetime
 
 import django
+
+# Use pip-vendored tomli so we can read pyproject.toml
+# When we upgrade Python to 3.11 we can use tomllib directly
+from pip._vendor import tomli
 
 
 sys.path.insert(0, os.path.abspath(".."))  # for discovery of project modules
@@ -26,11 +32,13 @@ django.setup()
 # -- Project information -----------------------------------------------------
 
 project = "xocto"
-copyright = "2023, Kraken Tech"
+copyright = f"{datetime.now().year}, Kraken Tech"
 author = "Kraken Tech"
 
-# The full version, including alpha/beta/rc tags
-release = "2.3.0"
+# Fetch the version from pyproject.toml
+path = pathlib.Path(__file__).parent / ".." / "pyproject.toml"
+pyproject = tomli.loads(path.read_text())
+release = pyproject["project"]["version"]
 
 # -- General configuration ---------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -26,11 +32,13 @@
 # -- Project information -----------------------------------------------------
 
 project = "xocto"
-copyright = "2023, Kraken Tech"
+copyright = f"{datetime.now().year}, Kraken Tech"
```
jarshwah, line 35: 🐼  I've always wondered.. should we use the current year or the year of first publication? Most things I'm reading at the moment suggest the original year is better, but overall it doesn't matter 🙃

docs/conf.py
```diff
@@ -26,11 +32,13 @@
...
 author = "Kraken Tech"
 
-# The full version, including alpha/beta/rc tags
-release = "2.3.0"
+# Fetch the version from pyproject.toml
+path = pathlib.Path(__file__).parent / ".." / "pyproject.toml"
+pyproject = tomli.loads(path.read_text())
+release = pyproject['project']['version']
```
jarshwah, line 41: Nice!

pyproject.toml
```diff
@@ -63,10 +63,10 @@ dev = [
...
 ]
 docs = [
-  "Sphinx==4.5.0",
-  "myst-parser==0.18.1",
-  "sphinxcontrib-serializinghtml==1.1.5",
-  "sphinx-rtd-theme==1.1.1",
+  "Sphinx==7.2.6",
+  "myst-parser==2.0.0",
+  "sphinxcontrib-serializinghtml==1.1.10",
+  "sphinx-rtd-theme==2.0.0",
```
jarshwah, line 69: You're a brave man. I attempted all kinds of version mixes and couldn't get `make html` in the docs directory to work without spewing errors. Though, after I messed around some, I suspect they were probably just warnings (which persist prior to this PR). Have you confirmed it builds?

jarshwah (approved):
Thank you 👍 

I notice the build is failing (ruff check) - the pre-commit should be fairly good now (except for https://github.com/octoenergy/xocto/pull/136/files#r1508516144 which I'd be happy for you to bring across here if you use pre-commit)

pyproject.toml
pydanny: Builds locally. I see we don't have branch builds active in RTD, seeing how to turn that on.

docs/conf.py
pydanny: One version change to rule them all! 😁

PR #142, opened 2024-03-05
Make `Range` instances somewhat immutable

quicklizard99:
This PR addresses two problems with `Range`.

1. Attributes of `Range` instances can be set, bypassing validation and violating class invariants:

    ```pycon
    >>> from xocto import ranges
    >>> r = ranges.Range(0, 1)
    >>> r.start = 2
    >>> r.boundaries = '😱'
    ```

2. Additional attributes can be set on instances of classes derived from `Range`

    `Range` uses `__slots__`, which prevents creation of instance dictionaries. Classes derived from `Range` do not have `__slots__ = ()`, which means that instance dictionaries are created:

    ```pycon
    >>> from xocto import localtime
    >>> r = ranges.FiniteDatetimeRange(localtime.datetime(2000, 1, 1), localtime.datetime(2000, 1, 3))
    >>> r.some_other_thing = "I shouldn't be here"
    >>> r.__dict__
    {'some_other_thing': "I shouldn't be here"}
    ```

Fix these problems:

* add `__slots__ = ()` to classes that have `Range` as a base class (preventing creation of instance dicts),
* add `Range.__setattr__` to raise `AttributeError`, and
* alter `Range.__init__` to use the same `object.__setattr__` [trick as `attrs`](https://www.attrs.org/en/stable/init.html#post-init) to set attributes during initialisation.

### DIFF
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -143,7 +143,7 @@ class Range(Generic[T]):
         None
     """
 
-    __slots__ = [
+    __slots__ = (
         "start",
         "end",
         "boundaries",
@@ -151,7 +151,15 @@ class Range(Generic[T]):
         "_is_left_inclusive",
         "_is_right_exclusive",
         "_is_right_inclusive",
-    ]
+    )
+
+    start: Optional[T]
+    end: Optional[T]
+    boundaries: RangeBoundaries
+    _is_left_exclusive: bool
+    _is_left_inclusive: bool
+    _is_right_exclusive: bool
+    _is_right_inclusive: bool
 
     def __init__(
         self,
@@ -448,6 +456,9 @@ class FiniteRange(Range[T]):
     and then skip checking if the endpoints are None.
     """
 
+    __slots__ = ()
+
+    # Redefine types in base class
     start: T
     end: T
 
@@ -473,8 +484,13 @@ class HalfFiniteRange(Range[T]):
     HalfFiniteRange.
     """
 
+    __slots__ = ()
+
+    # Redefine types in base class
     start: T
-    boundaries = RangeBoundaries.INCLUSIVE_EXCLUSIVE
+
+    def __init__(self, start: T, end: Optional[T] = None):
+        super().__init__(start, end, boundaries=RangeBoundaries.INCLUSIVE_EXCLUSIVE)
 
     def intersection(self, other: Range[T]) -> Optional["HalfFiniteRange[T]"]:
         """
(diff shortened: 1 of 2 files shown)


xocto/ranges.py
```diff
@@ -143,15 +143,23 @@ class Range(Generic[T]):
...
-    ]
+    )
+
+    start: Optional[T]
+    end: Optional[T]
+    boundaries: RangeBoundaries
+    _is_left_exclusive: bool
+    _is_left_inclusive: bool
+    _is_right_exclusive: bool
+    _is_right_inclusive: bool
```
quicklizard99, line 163: These types were previously implicitly declared when `__init__` set attributes.

xocto/ranges.py
```diff
@@ -473,8 +484,13 @@ class HalfFiniteRange(Range[T]):
     HalfFiniteRange.
     """
 
+    __slots__ = ()
+
+    # Redefine types in base class
     start: T
-    boundaries = RangeBoundaries.INCLUSIVE_EXCLUSIVE
```
quicklizard99, line 477: Setting the default value using a class attribute does not play with `__slots__`. I have moved this default to `__init__`.

xocto/ranges.py
```diff
@@ -229,6 +236,14 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"<Range: {str(self)}>"
 
+    def __setattr__(self, name: str, value: Any) -> None:
+        if name in type(self).__slots__:
+            raise AttributeError("Can't set attributes")
+        else:
+            raise AttributeError(
+                f"'{type(self).__name__}' object has no attribute '{name}'"
+            )
```
quicklizard99, line 245: These exception messages consistent with those raised by `collections.namedtuple`.

xocto/ranges.py
```diff
@@ -221,6 +236,14 @@ def __str__(self) -> str:
...
```
jamesbeith, line 245: Perhaps explicit is better than implicit, but I wonder if we could delegate this branch to `super()`?

jamesbeith, line 245: Perhaps explicit is better than implicit, but I wonder if we could delegate this branch to `super()`?

jarshwah (approved):
Good change, thanks for making it. As this is a potentially breaking change we'll need to at least bump the minor version when releasing.

xocto/ranges.py
quicklizard99: That is a good point. Change to use `super`

PR #144, opened 2024-03-12
Split xocto.types into a separate package

jarshwah:
Prior to this change, all types were in a single file, including the Django specific types that depend on Django itself.

This change creates a new types sub-package and splits it into separate modules. All of the types that depend on Django are now in a separate module.

Now xocto.ranges and xocto.numbers can be imported without having to install and configure Django.

As this is a breaking change, it'll need a major version number update.

### DIFF
--- a/xocto/types/generic.py
+++ b/xocto/types/generic.py
@@ -1,0 +1,21 @@
+from typing import Any, Protocol, TypeVar
+
+
+T = TypeVar("T", covariant=True)
+
+
+class Comparable(Protocol[T]):
+    """
+    Just a very basic way of describing an object that can be compared to another using some
+    basic operations.
+    """
+
+    def __eq__(self, other: Any) -> bool: ...
+
+    def __lt__(self, other: Any) -> bool: ...
+
+    def __le__(self, other: Any) -> bool: ...
+
+    def __gt__(self, other: Any) -> bool: ...
+
+    def __ge__(self, other: Any) -> bool: ...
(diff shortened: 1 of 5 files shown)


jamesbeith (approved):

xocto/types/generic.py
```diff
@@ -0,0 +1,21 @@
+from typing import Any, Protocol, TypeVar
+
+
+T = TypeVar("T", covariant=True)
+
+
+class Comparable(Protocol[T]):
+    """
+    Just a very basic way of describing an object that can be compared to another using some
```
delfick, line 9:
```suggestion
    A very basic way of describing an object that can be compared to another using some
```

irrelevant to your change.... but....

delfick (approved):

xocto/types/generic.py
jarshwah: Good call out - I considered making that change at the same time too ;)

jarshwah:
Will hold off merging this for a little while so 4.10 can stabilise, and allow any other minor/patch releases before going to v5 major.

xocto/types/generic.py
pydanny: You could get rid of `very` for a more concise sentence with the same meaning to readers. Very works better to the ear than it does to the eye.

pydanny (approved):
👍🏻

xocto/types/generic.py
jarshwah: I shortened it to "A way of.." which I think reads better 👍

Issue #149, opened 2024-03-28
Range fields should accept scalar values when querying


jarshwah:
https://github.com/octoenergy/xocto/commit/93c37c9055d7cd545d2cbca1c41a5792cbe3b15b assumes that a range should always be passed when querying a range field, but that's not the case, particularly when performing a `contains` query:

```
Model.objects.filter(period__contains=datetime.datetime(2024, 3, 1))
# TypeError(FiniteDateTimeRangeField may only accept FiniteDatetimeRange objects)
```

We're too aggressive with the type checking. Django fields are a lot less strict, allowing ranges, tuples, and scalar values: https://github.com/django/django/blob/944745afe2ec45aed30cef799c250107f1364ca7/django/contrib/postgres/fields/ranges.py#L89-L96

We should just follow along, but perhaps prevent the case of using tuples, because the bounds are not well defined in that case.

PR #150, opened 2024-03-28
Allow passing single date/datetime to postgres db range fields

Peter554:
Fixes https://github.com/octoenergy/xocto/issues/149

### DIFF
--- a/tests/fields/postgres/test_ranges.py
+++ b/tests/fields/postgres/test_ranges.py
@@ -2,6 +2,7 @@ import datetime
 import zoneinfo
 
 import pytest
+from dateutil import relativedelta
 from django.conf import settings
 from django.core import serializers
 from django.utils import timezone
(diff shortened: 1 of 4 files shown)


tests/storage/test_storage.py
```diff
@@ -1226,7 +1226,7 @@ def test_query_dataframe_with_sql(self):
         assert result_df.shape == (1, 2)
 
         assert result_df["string_column"][0] == "A"
-        assert result_df["array_column"][0] == [1, 2, 3]
+        assert tuple(result_df["array_column"][0]) == (1, 2, 3,)
```
Peter554, line 1229: I don't understand why this test was failing, and this isn't a fix that suggests any understanding, but it does work.

tests/fields/postgres/test_ranges.py
```diff
@@ -283,36 +340,66 @@ def test_query(self):
...
                 half_finite_datetime_range__overlap=(
                     localtime.datetime(2024, 1, 1),
                     localtime.datetime(2024, 1, 15),
                 )
             )
-        with pytest.raises(TypeError):
+        with pytest.raises(
+            TypeError,
+            match="HalfFiniteDateTimeRangeField may only accept HalfFiniteDateTimeRangeField or datetime objects",
+        ):
```
AzinKhan, line 382: 🐼 The use of `match=` is a good addition but perhaps it warrants a separate commit?

xocto/fields/postgres/ranges.py
```diff
@@ -25,15 +25,21 @@ class FiniteDateRangeField(pg_fields.DateRangeField):
...
-        if not isinstance(value, ranges.FiniteDateRange):
-            raise TypeError(
-                "FiniteDateRangeField may only accept FiniteDateRange objects."
-            )
-        return pg_ranges.DateRange(lower=value.start, upper=value.end, bounds="[]")
+        if isinstance(value, ranges.FiniteDateRange):
+            return pg_ranges.DateRange(lower=value.start, upper=value.end, bounds="[]")
+        if (
+            isinstance(value, datetime.date)
+            # Don't allow datetime (datetime is a subclass of date!)
```
AzinKhan, line 36: This has always bugged me about the datetime library 😐 .

xocto/fields/postgres/ranges.py
```diff
@@ -158,22 +166,18 @@ class HalfFiniteDateTimeRangeField(pg_fields.DateTimeRangeField):
     """
 
     def get_prep_value(
-        self, value: Optional[ranges.HalfFiniteDatetimeRange]
-    ) -> Optional[pg_ranges.DateTimeTZRange]:
+        self, value: Any
```
AzinKhan, line 169: Why `Any` here and not `Optional[Union[ranges.HalfFiniteDatetimeRange, datetime.datetime]]`? (And similar for the other `get_prep_value` implementations).

Peter554: By my understanding, `Any` will be a more correct type hint here. We can't enforce that the value will be `Optional[Union[ranges.HalfFiniteDatetimeRange, datetime.datetime]]`, since there is nothing stopping us doing things like `MyModel.objects.filter(period__contains="foo")` (or any other arbitrary type). This type checking happens at runtime (hence the `raise TypeError`), not mypy.

tests/fields/postgres/test_ranges.py
Peter554: Okay I'll split this change out.

Peter554: Done

AzinKhan (approved):

PR #155, opened 2024-06-17
Add `union` method to `FiniteDatetimeRange`

GlennS:
Prior to this change, unioning two `FiniteDatetimeRange`s together would give you back a simple `Range[Datetime] | None`.

This change mirrors the implementation in `FiniteDateRange`.


jarshwah (approved):
Thanks

PR #160, opened 2024-06-18
Implement the copy and deepcopy protocol for Range

jarshwah:
Now that ranges are immutable, attempting to copy a range type would throw an error. This change implements the copy protocols by returning `self` directly. Since these are immutable instances there is no need to create a new copy. This is consistent with python itself, which will return functions directly rather than attempt to copy them.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import copy
 import datetime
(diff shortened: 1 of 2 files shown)


cogat (approved):
Looks good given the `setattr` and https://github.com/octoenergy/xocto/pull/142 enforcing immutability.

PR #163, opened 2024-06-19
Add ranges.iterate_over_months function

con-cat:
This commit adds a function to the ranges module that takes a finite date range and yields datetime ranges aligned to month boundaries and a supplied timezone during a period.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -9,7 +9,7 @@ import pytest
 from hypothesis import assume, given
 from hypothesis.strategies import composite, integers, none, one_of, sampled_from
 
-from xocto import ranges
+from xocto import localtime, ranges
 
 
 @composite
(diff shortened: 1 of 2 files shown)


xocto/ranges.py
```diff
@@ -1034,3 +1038,34 @@ def as_finite_datetime_periods(
         finite_periods += [FiniteDatetimeRange(period.start, period.end)]
 
     return finite_periods
+
+
+def iterate_over_months(
+    period: FiniteDateRange, tz: datetime.tzinfo
```
GlennS, line 1044: :panda_face: can I suggest making `tz` a named argument instead of positional?

xocto/ranges.py
```diff
@@ -1034,3 +1038,34 @@ def as_finite_datetime_periods(
...
+    ie: given a period of (15/01, 15/03) -> [
+        (15/01->01/02),
+        (01/02->01/03),
+        (01/03->15/03)
+    ]
+
+    Datetimes will be aligned to midnight as at the supplied timezone.
+    """
+    start_at = localtime.midnight(period.start, tz=tz)
+    end_at = localtime.next_midnight(period.end, tz=tz)
```
GlennS, line 1060: What happens if `end_at` is already midnight?

xocto/ranges.py
con-cat: I've made that update.

xocto/ranges.py
con-cat: I had made the `period` argument a `DateRange` instead of a `DatetimeRange` since we're rounding to midnight anyway, but thinking about it further that would then probably force people to convert datetimes to dates when calling this, which is annoying.

I've now changed `period` to be a `DatetimeRange`, and changed this line to `end_at = localtime.next_midnight(period.end, tz=tz)`

con-cat:
I've also added a few more test cases.

tests/test_ranges.py
```diff
@@ -996,6 +996,128 @@ def test_copies(self, obj):
         assert obj == copy.copy(obj) == copy.deepcopy(obj)
 
 
+class TestIterateOverMonths:
+    @pytest.mark.parametrize(
+        "row",
+        [
+            {
+                "period": ranges.FiniteDatetimeRange(
```
GlennS, line 1004: :panda_face: could we have ids for these tests?

GlennS (approved):
This looks correct to me. :+1: 

I spent some time trying to think of ways it might be broken that aren't covered by the implementation, and couldn't think of any.

xocto/ranges.py
GlennS: I think this is safe now.

xocto/ranges.py
```diff
@@ -1034,3 +1038,34 @@ def as_finite_datetime_periods(
...
+
+    Ranges span each whole month between the start and end times, inclusive.
+
+    ie: given a period of (15/01, 15/03) -> [
+        (15/01->01/02),
+        (01/02->01/03),
+        (01/03->15/03)
+    ]
+
+    Datetimes will be aligned to midnight as at the supplied timezone.
```
GlennS, line 1057: :panda_face: do we gain anything from doing this midnight alignment? What happens if we drop it entirely?

con-cat: I think we probably don't gain anything from this. The boundaries of the ranges will all be relative to the start datetime, but it would be just as easy and more flexible if we left it to the caller to do the midnight alignment if it's needed.

con-cat: I've removed the midnight alignment in 768be2148b8c65b218bc0da540efcdfaba57bf45

tests/test_ranges.py
con-cat: I've added ids in 768be2148b8c65b218bc0da540efcdfaba57bf45

PR #166, opened 2024-07-22
Rename pact_service parameters

jnns:
A pact version refers to the version of a pact file and is handled by the pact framework internally. Referring to the application version reference as `pact_version` can create confusion. For a long time, [we defaulted to a consumer version of `latest` in kraken-core](https://github.com/octoenergy/kraken-core/commit/66130696bfe353ab3a09a316ce4c03747248025e#diff-d693b2faa093742e7045e9d6606e499d5ff59a8c9625968286d5abf230c28e35L1262) - which should not be the case <sup>[1](https://docs.pact.io/getting_started/versioning_in_the_pact_broker), [2](https://docs.pact.io/pact_broker/pacticipant_version_numbers)</sup>. 

- This pull request changes the name of the parameter `pact_version` to `consumer_version`. 
- The `pact_` prefix of all the other parameter names is dropped as well because Pact is the whole context of that function and there doesn't need to be a differentiation to other parameters.

**Extra**:

As this is the first time for me contributing to this package and I just installed the requirements today, Django 5 was installed instead of Django 4. Running the tests led to an error due to non-existing deprecation warnings that were ignored previously. I fixed the test and removed the ignore directives.

### DIFF
--- a/xocto/pact_testing.py
+++ b/xocto/pact_testing.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import dataclasses
 import json
 import os
 import subprocess
@@ -7,6 +8,17 @@ from typing import Any
 
 import pact
 import requests
+
+
+@dataclasses.dataclass(frozen=True)  # Use kw_only=True once we're on Python 3.10.
+class PactOptions:
+    broker_url: str
+    broker_username: str
+    broker_password: str
+    consumer_name: str
+    provider_name: str
+    consumer_version: str
+    log_path: str
 
 
 class PactConsumerClient:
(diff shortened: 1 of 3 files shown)


xocto/pact_testing.py
```diff
@@ -30,25 +30,25 @@ def post(
 
 
 def pact_service(
-    pact_broker_url: str,
-    pact_broker_username: str,
-    pact_broker_password: str,
-    pact_consumer_name: str,
-    pact_provider_name: str,
-    pact_consumer_version: str,
+    broker_url: str,
```
delfick, line 33: It feels like we should instead do something like

```
@dataclasses.dataclass(frozen=True, kw_only=True)
class PactOptions:
    broker_username: str
    broker_password: str
    consumer_name: str
    provider_name: str
    consumer_version: str
    log_path: str

def pact_service(*, options: PactOptions, publish_to_broker: bool) -> pact.Pact:
```

delfick (commented):
Seems reasonable. We should probably split the django fix into a separate PR too.

jnns:
Thank you for your feedback @delfick!

I grouped the passed data into a parameter object. Sadly, we cannot use `kw_only=True` until we're on Python 3.10. 

I also removed the commit for compatibility with Django 5.

xocto/pact_testing.py
```diff
@@ -9,6 +10,17 @@
 import requests
 
 
+@dataclasses.dataclass(frozen=True)  # Use kw_only=True once we're on Python 3.10.
```
delfick, line 13: according to the docs kw_only is available since python3.10 so it should work here

delfick (approved):
nice

xocto/pact_testing.py
jnns: Our docs state we have compatibility with Python>=3.9. My train of thought was that we can only make use of `kw_only` once we no longer need to provide compatibility with Python 3.9. 🤔 

> This library works with Python 3.9 and above. ([source](https://xocto.readthedocs.io/en/latest/index.html))

https://github.com/octoenergy/xocto/blob/b808a5c769b95fd95a43155e44b938c0eb221dea/pyproject.toml#L8

delfick: oh, sorry, I misunderstood the comment.

PR #170, opened 2024-08-19
Make tests pass in Django>=5

jnns:
Django 5.0 removed the alias `django.utils.timezone.utc` to `datetime.timezone.utc`.

Remove unnecessary deprecation warning filters for pytest.


delfick (approved):
thanks

PR #171, opened 2024-08-29
Fix bug with `list_files`  method in `S3SubdirectoryFileStore`

DHUKK:
🐛 Fix bug with `list_files`  method in `S3SubdirectoryFileStore`

Previously
---
If `S3SubdirectoryFileStore` was initialized with no path other than the s3 bucket
ie. `s3://bucket-name`
The `list_files` method would incorrectly trim the returned paths

e.g
expected: 
```
something/other/a.txt
```
returned: 
```
omething/other/a.txt
```

This change
---
Does not trim the returned paths if `S3SubdirectoryFileStore` is initialized with no path.

### DIFF
--- a/tests/storage/test_storage.py
+++ b/tests/storage/test_storage.py
@@ -175,7 +175,6 @@ class TestS3SubdirectoryFileStore:
             Prefix="folder/a/b"
         )
 
-    @pytest.mark.xfail(reason="Expose bug when listing files with no bucket path")
     @mock.patch.object(storage.S3FileStore, "_get_boto_bucket")
     def test_list_files_no_path(self, get_boto_bucket):
         get_boto_bucket.return_value = mock.Mock(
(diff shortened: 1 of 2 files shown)


mmxw (approved):
Thank you so much for fixing it!

tests/storage/test_storage.py
```diff
@@ -175,6 +175,28 @@ def test_list_files(self, get_boto_bucket):
...
+                "objects.filter.return_value": [
+                    mock.Mock(key="folder/a/b/"),
+                    mock.Mock(key="folder/a/b/foo.txt"),
+                    mock.Mock(key="folder/a/b/bar.txt"),
+                ]
+            }
+        )
+        store = storage.S3SubdirectoryFileStore("s3://some-bucket")
+
+        assert list(store.list_files(namespace="folder/a/b")) == [
```
mmxw, line 191: Just to confirm - I can safely assume that I don't need a trailing for prefixing `/` for the namespace, right?

DHUKK: Yea you shouldn't need a trailing `/` for the namespace, that was a wrong assumption I made initially

noelia-mari (approved):
Thanks for this!

PR #173, opened 2024-09-09
Update date_of_day_before docstring

federicodabrunzo:
Prior to this change, the docstring was referring to an upper bound inclusive convention for date ranges. Since there is no such convention in the Kraken coding conventions, this can be confusing for the developers.

This change updates the docstring to reflect the fact that it is a thing we usually do, but there's no written convention for it.


jbkkd:
Closing as stale for now, reopen if you still want to merge this

PR #174, opened 2024-10-09
Add parameter to DateTime-based range fields

tdcdev:
## Prior to this change

Since https://github.com/octoenergy/xocto/commit/7558f7105dc4053bca190e927df0206923ea30a9 values returned from DB are converted in local timezone.

## This change

This change adds an optional parameter allowing to choose whether or not to convert values into local timezone.

## Why it's needed

We cannot use those fields with the last hour of the DST period (demonstration here: #169). We need an option to not convert values to local timezone and keep them in UTC.

### DIFF
--- a/xocto/fields/postgres/ranges.py
+++ b/xocto/fields/postgres/ranges.py
@@ -93,7 +93,18 @@ class FiniteDateRangeField(pg_fields.DateRangeField):
         return value.upper - datetime.timedelta(days=1)
 
 
-class FiniteDateTimeRangeField(pg_fields.DateTimeRangeField):
+class _LocaliserMixin:
+    def __init__(
+        self, *args: Any, timezone: Optional[datetime.tzinfo] = None, **kwargs: Any
+    ):
+        super().__init__(*args, **kwargs)
+        self._timezone = timezone
+
+    def localise(self, value: datetime.datetime) -> datetime.datetime:
+        return localtime.as_localtime(value, self._timezone)
+
+
+class FiniteDateTimeRangeField(_LocaliserMixin, pg_fields.DateTimeRangeField):
     """
     A DateTimeRangeField with Inclusive-Exclusive [) bounds that aren't infinite.
 
@@ -125,8 +136,8 @@ class FiniteDateTimeRangeField(pg_fields.DateTimeRangeField):
         if value is None:
             return None
         return ranges.FiniteDatetimeRange(
-            start=localtime.as_localtime(value.lower),
-            end=localtime.as_localtime(value.upper),
+            start=self.localise(value.lower),
+            end=self.localise(value.upper),
         )
 
     def to_python(self, value: Optional[str]) -> Optional[ranges.FiniteDatetimeRange]:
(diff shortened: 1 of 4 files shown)


xocto/fields/postgres/ranges.py
```diff
@@ -93,7 +93,20 @@ def _upper_to_inclusive(self, value: pg_ranges.DateRange) -> datetime.date:
...
-class FiniteDateTimeRangeField(pg_fields.DateTimeRangeField):
+class _ConvertibleDateTimeRangeMixin:
+    def __init__(self, *args: Any, as_localtime: bool = True, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self._as_localtime = as_localtime
+
+    def convert(self, value: datetime.datetime) -> datetime.datetime:
+        if self._as_localtime:
+            return localtime.as_localtime(value)
+        return localtime.as_utc(value)
```
tdcdev, line 104: ❓ I'm not comfortable with this naming:
* `_ConvertibleDateTimeRangeMixin`
* `as_localtime`
* `convert`

 Are you thinking of something better ?

tdcdev (commented):
I haven't updated the documentation yet as I'm not sure about the naming.

xocto/fields/postgres/ranges.py
```diff
@@ -93,7 +93,20 @@ def _upper_to_inclusive(self, value: pg_ranges.DateRange) -> datetime.date:
         return value.upper - datetime.timedelta(days=1)
 
 
-class FiniteDateTimeRangeField(pg_fields.DateTimeRangeField):
+class _ConvertibleDateTimeRangeMixin:
+    def __init__(self, *args: Any, as_localtime: bool = True, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self._as_localtime = as_localtime
```
jarshwah, line 99: **suggestion**: rather than a boolean, should we consider having a timezone argument? The "default" can be the localtime timezone - but you can also pass in UTC.

That way we can progressively update callers to using UTC?

That said it's probably unlikely callers would care about anything other than LOCAL and UTC right? So that might not be better..

xocto/fields/postgres/ranges.py
jarshwah: `_Localiser`
`localise`

?

jarshwah (commented):
I think this is probably very close to correct. Have left a couple of comments/questions, and will approve once answered/addressed.

xocto/fields/postgres/ranges.py
tdcdev: Yes I think it's a great idea! This allows more flexibility if the user wants.

> That said it's probably unlikely callers would care about anything other than LOCAL and UTC right?

I agree, but with hindsight it seems a bit odd to me adding a boolean only for UTC/LOCAL conversion as `as_localtime` allows to choose a timezone.

xocto/fields/postgres/ranges.py
tdcdev: Much better thanks!

jarshwah (approved):
Thanks, this is good 👍 

Recommend adding a test that demonstrates that the issue fails with localtime but passes with utc as per https://github.com/octoenergy/xocto/pull/169 but that's non-blocking.

PR #175, opened 2024-10-15
Parse str finite date ranges

DanielOwen1:


xocto/ranges.py
```diff
@@ -965,6 +965,32 @@ def days(self) -> int:
         return (self.end - self.start).days + 1
 
 
+def finite_date_range_from_str(period_str: str) -> FiniteDateRange:
```
benthorner, line 968: ❓ How about making this a class method of `Range`?

```python
Range.parse(...)
```

This is a little more complicated, but makes sense to me as a complement to the existing `__str__` method, which renders the same format:

https://github.com/octoenergy/xocto/blob/49a1a8313a4e9b50b4f504f70ca192156f66f82c/xocto/ranges.py#L227-L238

benthorner (commented):
Being able to "parse" a period makes sense to me ➕.

I think we need to do a little more validation to ensure what we're parsing is actually period-like, rather than (just) an ambiguous CSV of dates.

xocto/ranges.py
```diff
@@ -954,6 +954,31 @@ def is_disjoint(self, other: Range[datetime.date]) -> bool:
...
+        Coverts a str into a FiniteDateRange.
+
+        :raises ValueError:
+            If the period string has the incorrect number of parts.
+            If the period is not finite.
+
+        >>> FiniteDateRange.parse("[2021-01-01,2021-01-02]")
+        FiniteDateRange(start=datetime.date(2021, 1, 1), end=datetime.date(2021, 1, 2))
+        """
+        parts = period_str.strip(RangeBoundaries.INCLUSIVE_INCLUSIVE.value).split(",")
```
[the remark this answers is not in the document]
DanielOwen1: Good point. I've amended to check for the correct boundaries.

xocto/ranges.py
```diff
@@ -954,6 +954,37 @@ def is_disjoint(self, other: Range[datetime.date]) -> bool:
...
+
+        :raises ValueError:
+            If the range string has incorrect boundaries.
+            If the range string has an incorrect number of parts.
+            If the range is not finite.
+
+        >>> FiniteDateRange.parse("[2021-01-01,2021-01-02]")
+        FiniteDateRange(start=datetime.date(2021, 1, 1), end=datetime.date(2021, 1, 2))
+        """
+        left_bracket = range_str[0]
```
benthorner, line 970: 🐼 This could raise an `IndexError` for an empty string.

benthorner (approved):
Thanks for addressing my comments 🏅.

Looks good to go from my perspective, but we should get someone who's not been in previous discussions about this to give an opinion on it as a feature.

xocto/ranges.py
```diff
@@ -954,6 +954,37 @@ def is_disjoint(self, other: Range[datetime.date]) -> bool:
             Range(start=other_start, end=other_end, boundaries=other.boundaries)
         )
 
+    @classmethod
+    def parse(cls, range_str: str) -> FiniteDateRange:
```
jarshwah, line 958: I'm not sure this belongs here to be honest.

Either parsing should be supported for all range types - or serialization/deserialization belongs elsewhere. I'd probably lean towards the latter..

jarshwah (commented):
I don't think the range types themselves should embed serialization/deserialization - that belongs elsewhere.

Not blocking merge because this is really opinion based..

xocto/ranges.py
benthorner: > Either parsing should be supported for all range types

Ideally yes, but that's quite a complex problem: to parse arbitrary range boundaries from strings.

> or serialization/deserialization belongs elsewhere

What's the negative impact of having a standard way of deserializing date ranges specifically? 

I'm thinking about this less as de/serialization and more as a kind of pseudo-literal syntax for ranges: it's a shame Python doesn't support ranges as first class literals.

jarshwah: I’m thinking it probably doesn’t belong in this library at all

PR #177, opened 2024-11-07
Fix local storage fetch_url

WillC3:
## Description
Fixes issue where if a local url is passed with the bucket_name in, `fetch_url` will produce a url with an extra `/`, eg `/media-url/bucket-name//some/key`.

### DIFF
--- a/tests/storage/test_storage.py
+++ b/tests/storage/test_storage.py
@@ -1003,13 +1003,13 @@ class TestLocalFileStore:
         fetch_url = store.fetch_url("some/key", version_id="some-version")
         assert fetch_url == "/media-url/bucket-name/some/some-version/key"
 
-    @pytest.mark.xfail
     @override_settings(MEDIA_ROOT="/media-root/", MEDIA_URL="/media-url/")
     def test_fetch_url_with_key_path_containing_root_and_bucket(self):
         store = storage.LocalFileStore("bucket-name")
         fetch_url = store.fetch_url("/media-root/bucket-name/some/key")
         assert fetch_url == "/media-url/bucket-name/some/key"
 
+    @pytest.mark.xfail
     @override_settings(MEDIA_ROOT="/media-root/", MEDIA_URL="/media-url/")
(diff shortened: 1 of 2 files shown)


tests/storage/test_storage.py
```diff
@@ -1003,6 +1003,12 @@ def test_fetch_url_with_version(self):
         fetch_url = store.fetch_url("some/key", version_id="some-version")
         assert fetch_url == "/media-url/bucket-name/some/some-version/key"
 
+    @override_settings(MEDIA_ROOT="/media-root/", MEDIA_URL="/media-url/")
+    def test_fetch_url_with_key_path_containing_root_and_bucket(self):
+        store = storage.LocalFileStore("bucket-name")
+        fetch_url = store.fetch_url("/media-root/bucket-name/some/key")
```
alexeocto, line 1009: would you not use the setting here rather than hard code the media root into the arg?

alexeocto: also I don't think we are passing it the bucket in the url key path are we?

tests/storage/test_storage.py
```diff
@@ -1003,6 +1003,12 @@ def test_fetch_url_with_version(self):
...
+        assert fetch_url == "/media-url/bucket-name/some/key"
```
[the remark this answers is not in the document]
WillC3: Gonna be honest I don't really know what the difference is between `MEDIA_ROOT` & `MEDIA_URL`. I was only really trying to fix this small bug with the PR, not change functionality

alexeocto (approved):
Very strange one. Also obscured by the fact that the bug doesn't show up when actually using this with the local filestore in kraken in the UI cos the url pattern matching actually manages to overcome the extra /.

Whether this actually impacts anything outside of tests in kraken core that validate the returned values from some methods im not even sure. 

Should probably get someone else to double check this as well before merging and that all the versioning stuff is correct.

WillC3:
I think considering this isn't really affecting anything I'm happy to just close for now

PR #178, opened 2024-11-22
Add "localize" and "date_range_for_midnight_range" fns

benthorner:
Follow-up to [^2].

Together these can be used to DRY-up converting datetime ranges to 
date ranges, irrespective of how the datetime range has been stored.

Before:

    ranges.FiniteDateRange(
        localtime.date(midnight_aligned_range.start),
        localtime.date_of_day_before(midnight_aligned_range.end),
    )

After:

    ranges.date_range_for_midnight_range(
        midnight_aligned_range.localize(timezone.get_current_timezone()),
    )

Note: in the example the datetime range should be midnight-aligned 
in the target timezone, which may or may not be its current timezone.

Pull Request in Kraken Core to demo usage [^1].

[^1]: https://github.com/octoenergy/kraken-core/pull/166268
[^2]: https://github.com/octoenergy/kraken-core/pull/166008#discussion_r1842744663

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1187,6 +1187,54 @@ class TestIterateOverMonths:
         assert result == row["expected"]
 
 
+class TestDateRangeForMidnightRange:
+    def test_returns_date_range(self):
+        dt_range = ranges.FiniteDatetimeRange(
+            datetime.datetime(2020, 1, 1),
+            datetime.datetime(2020, 1, 10),
+        )
+
+        assert ranges.date_range_for_midnight_range(dt_range) == ranges.FiniteDateRange(
+            datetime.date(2020, 1, 1),
+            datetime.date(2020, 1, 9),
+        )
+
+    def test_errors_if_different_timezones(self):
+        dt_range = ranges.FiniteDatetimeRange(
+            datetime.datetime(2020, 1, 1, tzinfo=zoneinfo.ZoneInfo("Asia/Dubai")),
+            datetime.datetime(
+                2020, 1, 10, tzinfo=zoneinfo.ZoneInfo("Australia/Sydney")
+            ),
+        )
+
+        with pytest.raises(ValueError) as exc_info:
+            ranges.date_range_for_midnight_range(dt_range)
+
+        assert "Start and end in different timezones" in str(exc_info.value)
+
+    def test_errors_if_start_not_midnight(self):
+        dt_range = ranges.FiniteDatetimeRange(
+            datetime.datetime(2020, 1, 1, hour=1),
+            datetime.datetime(2020, 1, 10),
+        )
+
+        with pytest.raises(ValueError) as exc_info:
+            ranges.date_range_for_midnight_range(dt_range)
+
+        assert "Start of range is not midnight-aligned" in str(exc_info.value)
+
+    def test_errors_if_end_not_midnight(self):
+        dt_range = ranges.FiniteDatetimeRange(
+            datetime.datetime(2020, 1, 1),
+            datetime.datetime(2020, 1, 10, hour=1),
+        )
+
+        with pytest.raises(ValueError) as exc_info:
+            ranges.date_range_for_midnight_range(dt_range)
+
+        assert "End of range is not midnight-aligned" in str(exc_info.value)
+
+
 def _rangeset_from_string(rangeset_str: str) -> ranges.RangeSet[int]:
     """
     Convenience method to make test declarations clearer.
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import copy
 import datetime
 import re
+import zoneinfo
 from typing import Any
 
 import pytest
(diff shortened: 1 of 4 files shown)


tests/test_ranges.py
```diff
@@ -318,6 +318,34 @@ def test_range_difference_and_intersection_form_partition(
                 # Ignore types here as structuring this to appease mypy would make it v ugly.
                 assert (a_difference | intersection | b_difference) == (a | b)  # type: ignore[operator]
 
+    class TestCopy:
```
benthorner, line 322: 🧀 Moved up / renamed from standalone test class below 👇.

xocto/ranges.py
```diff
@@ -886,6 +886,24 @@ def seconds(self) -> int:
...
+        """
+        Returns the range with boundaries adjusted to the specified timezone.
+
+        See datetime.astimezone for more details.
+
+        Raises:
+            ValueError:
+                If one or both boundaries are naive (no timezone).
+        """
+        if not self.start.tzinfo or not self.end.tzinfo:
```
benthorner, line 899: 🧀 `astimezone` will assume the system timezone by default [^1], which may or may not be correct - best to error, which is consistent with the behaviour for `localtime` / Django.

[^1]: https://docs.python.org/3/library/datetime.html#datetime.datetime.astimezone

tests/test_ranges.py
```diff
@@ -964,6 +965,34 @@ def test_union_of_overlapping_ranges(self):
                 end=datetime.datetime(2000, 1, 4),
             )
 
+    class TestLocalize:
+        def test_converts_timezone(self):
```
jarshwah, line 969: Can you please add a test here over DST boundaries? Specifically what happens when you have an hour range (in utc) and the start localises to a different tz-offset, and a separate test when end localises to a different tz-offset.

I'm not confident `astimezone` will do the right thing - whether that correctly adjusts the offset or not. We dealt with a similar issue of DateTimeRangeField (that would automatically localise) and it would result in empty ranges, because the post-localised value included a `fold=1` attribute indicating which *side* of the boundary it should be on, but the start and end values themselves had the same *time*.

xocto/ranges.py
```diff
@@ -886,6 +886,24 @@ def seconds(self) -> int:
         """
         return int((self.end - self.start).total_seconds())
 
+    def localize(self, tz: datetime.tzinfo) -> FiniteDatetimeRange:
```
jarshwah, line 889: Is `datetime.tzinfo` what we want, or should we be looking at ZoneInfo?

jarshwah (commented):
I think this is looking ok - a few questions left about timezones in general - will be happy to approve once comments are addressed.

jarshwah (changes requested):
One more blocking change - referencing `localtime` in `day_before` - once that's resolved I'm happy to approve.

tests/test_ranges.py
benthorner: Very interesting stuff!

> Localising a datetime, natively, does not work well with ranges

In what way?

Some periods of time just "don't exist" in other timezones - the errors reflect that.

> that should be 0 seconds - but the timedelta module doesn't consider offsets important

I'm not sure this example is realistic: 2AM doesn't actually exist at that point in time.

Arguably [`datetime` itself should error on creation if the point in time doesn't exist](https://peps.python.org/pep-0495/#the-fold-attribute):

> ...instances that fall in the spring-forward gap. Such instances don’t represent any valid time

But this isn't just [Garbage In, Garbage Out](https://en.wikipedia.org/wiki/Garbage_in,_garbage_out), as `timedelta` does emit invalid datetimes:

```python
datetime(2024, 10, 6, 2, tzinfo=SYD) + timedelta(hours=1)
=> datetime.datetime(2024, 10, 6, 2, 0, tzinfo=...)  # <== WRONG
```

Overall this is a problem with `datetime` itself. If it weren't a problem, I don't think we
would be having this discussion, or any qualms about this function.

🐮 For now I've done a PR to document this pitfall: https://github.com/octoenergy/xocto/pull/180

🐮 Maybe we should wrap `datetime` so it actually behaves correctly 🐉 💥.

xocto/ranges.py
```diff
@@ -1067,3 +1085,34 @@ def iterate_over_months(
...
+    # Check datetimes are both midnight-aligned.
+    if range.start.time() != datetime.time(0, 0):
+        raise ValueError("Start of range is not midnight-aligned")
+
+    if range.end.time() != datetime.time(0, 0):
+        raise ValueError("End of range is not midnight-aligned")
+
+    return FiniteDateRange(
+        range.start.date(),
+        localtime.day_before(range.end.date()),
```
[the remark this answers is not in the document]
benthorner: Oops, sorry about that: fixed in the (now) final commit.

tests/test_ranges.py
benthorner: (I've fixed up the new tests into the second commit, which adds the function under test.)

tests/test_ranges.py
```diff
@@ -964,6 +965,69 @@ def test_union_of_overlapping_ranges(self):
...
+            )
+
+        def test_errors_converting_over_dst_gain_hour(self):
+            utc_tz = zoneinfo.ZoneInfo("UTC")
+            london_tz = zoneinfo.ZoneInfo("Europe/London")
+
+            # Create a range in London over the hour that is "gained"
+            # when Daylight Savings Time (DST) starts - at 1AM.
+            #
+            # Note: this is allowed by datetime but not a realistic
```
benthorner, line 992: 🧀 Added comment that this test isn't very realistic.

jarshwah (approved):

tests/test_ranges.py
jarshwah: One more thing - the other side of DST where time goes backwards is similarly bad, and isn't an example of garbage in.

```
In [36]: localtime.as_localtime(datetime(2024, 4, 6, 17, tzinfo=UTC), tz=SYD) - localtime.as_localtime(datetime(2024, 4, 6, 16, tzinfo=UTC), tz=SYD)
Out[36]: datetime.timedelta(seconds=3600)

In [37]: localtime.as_localtime(datetime(2024, 4, 6, 17, tzinfo=UTC), tz=SYD) - localtime.as_localtime(datetime(2024, 4, 6, 15, tzinfo=UTC), tz=SYD)
Out[37]: datetime.timedelta(seconds=3600)
```

Here, we see two 2 hours of UTC being localised to 1 hour of duration, because resolving it to "SYD" is the 2am *after* the wind-back, and hence the fold attribute is set:

```
In [39]: localtime.as_localtime(datetime(2024, 4, 6, 16, tzinfo=UTC), tz=SYD)
Out[39]: datetime.datetime(2024, 4, 7, 2, 0, fold=1, tzinfo=zoneinfo.ZoneInfo(key='Australia/Sydney'))
```

Timezone *names* aren't helpful for comparing durations over DST boundaries. It's all.. annoying and hard.

benthorner: > Should dict[Any, datetime] support localize(tz=...)?

Well, no, but I'd say that's more because `dict` is generic, where `FiniteDatetimeRange` is specific.

> the other side of DST where time goes backwards is similarly bad

Hmm...thought provoking. To make the localization explicit:

- 2024-04-06 15:00 UTC => 2024-04-07 02:00 SYD
- 2024-04-06 16:00 UTC => 2024-04-07 02:00 SYD
- 2024-04-06 17:00 UTC => 2024-04-07 03:00 SYD

> Here, we see two 2 hours of UTC being localised to 1 hour of duration

Is it wrong though? SYD is losing an hour after all, so that has to have an effect somewhere.

jarshwah:
Good to merge @benthorner - though we usually only need a version bump immediately if there’s a major version shift (to ensure there aren’t releases of other minor changes). There’s not a lot of activity on this package though so it’s fine to bump now.

PR #181, opened 2024-12-18
Encapsulate date_range... fn in FiniteDatetimeRange

benthorner:
Follow-up to [^1].

This moves/renames `ranges.date_range_for_midnight_range` so it's
a method: `FiniteDatetimeRange.as_date_range` - more concise.

Before:

    ranges.date_range_for_midnight_range(
        dt_range.localize(my_timezone)  # common pre-step
    )

After:

    dt_range.localize(my_timezone).as_date_range()

The behaviour is unchanged: only midnight ranges, which fully span
a date range, are supported in the conversion.

[^1]: :pr: [Add "localize" and "date_range_for_midnight_range" fns](https://github.com/octoenergy/xocto/pull/178)

### DIFF
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -904,6 +904,36 @@ class FiniteDatetimeRange(FiniteRange[datetime.datetime]):
             self.end.astimezone(tz),
         )
 
+    def as_date_range(
+        self: FiniteDatetimeRange,
+    ) -> FiniteDateRange:
+        """
+        Returns the date range covered by this range (if midnight-aligned).
+
+        This can be useful where a range is available at datetime granularity,
+        but is used in functions that operate at date granularity.
+
+        Raises:
+            ValueError:
+                If the range boundaries are in different timezeones.
+                If the range boundaries are not midnight-aligned.
+        """
+        # First check range timezone is uniform.
+        if self.start.tzinfo != self.end.tzinfo:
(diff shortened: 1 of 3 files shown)


xocto/ranges.py
```diff
@@ -904,6 +904,36 @@ def localize(self, tz: datetime.tzinfo) -> FiniteDatetimeRange:
             self.end.astimezone(tz),
         )
 
+    def as_date_range(
+        range: FiniteDatetimeRange,
```
jarshwah, line 908: `self` ?

xocto/ranges.py
```diff
@@ -1085,34 +1115,3 @@ def iterate_over_months(
 
         yield FiniteDatetimeRange(start_at, this_end)
         start_at = next_start
-
-
-def date_range_for_midnight_range(
```
jarshwah, line 1090: If you want to avoid another major version / breaking change, consider changing the body of this to call the method?

jarshwah (changes requested):
One blocking change (self) one optional suggestion.

xocto/ranges.py
benthorner: I'd prefer to break: it's recent, so usages will very low if any (non in Kraken Core). Going through a full deprecation process doesn't seem worth it - but we can always change approach if someone complains.

PR #184, opened 2025-01-10
Optimise `ranges.any_overlapping`

Peter554:
This PR optimises the `ranges.any_overlapping` function. It uses `pytest-benchmark` to measure the improvement. The resulting function is >100x faster for this benchmark..!

Kraken: This function is often used to determine whether there are overlaps in meter readings etc and has a significant impact on costing performance.  

```
-------------------------------------------------------------------------------------------- benchmark: 2 tests --------------------------------------------------------------------------------------------
Name (time in ms)                            Min                 Max                Mean            StdDev              Median               IQR            Outliers       OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_any_overlapping (NOW)                1.4645 (1.0)        1.5542 (1.0)        1.4730 (1.0)      0.0099 (1.0)        1.4699 (1.0)      0.0032 (1.0)         70;93  678.8814 (1.0)         641           1
test_any_overlapping (0004_ce35b5e)     191.0346 (130.44)   207.6654 (133.61)   194.7505 (132.21)   6.6148 (666.40)   191.3241 (130.16)   4.8863 (>1000.0)       1;1    5.1348 (0.01)          6           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Improve the performance of the `ranges.any_overlapping` function
+  (with some benchmark showing a >100x speed up) [#184](https://github.com/octoenergy/xocto/pull/184).
+
 ## V7.0.0 - 2024-12-19
 
 - Add `FiniteDatetimeRange.as_date_range` [#181](https://github.com/octoenergy/xocto/pull/181)
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -1054,14 +1054,16 @@ def get_finite_datetime_ranges_from_timestamps(
 
 def any_overlapping(ranges: Iterable[Range[T]]) -> bool:
     """Return true if any of the passed Ranges are overlapping."""
-    ranges = list(ranges)
+    # We're deliberately not using RangeSet here for better performance.
+    # See https://github.com/octoenergy/xocto/pull/184.
     if not ranges:
         return False
-    range_set = RangeSet[T]([ranges[0]])
+    ranges = sorted(ranges)
+    prev_range: Range[T] = ranges[0]
     for range in ranges[1:]:
-        if range_set.intersection(range):
+        if prev_range.intersection(range):
             return True
-        range_set.add(range)
+        prev_range = range
     return False
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.benchmarks/
+
 # Byte-compiled / optimized files
 *.py[cod]
 __pycache__
--- a/makefile
+++ b/makefile
@@ -17,7 +17,7 @@ test:
 	py.test  --benchmark-skip
 
 benchmark:
-	py.test  --benchmark-only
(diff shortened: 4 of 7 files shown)


xocto/ranges.py
```diff
@@ -1054,14 +1054,15 @@ def get_finite_datetime_ranges_from_timestamps(
...
-    ranges = list(ranges)
-    if not ranges:
-        return False
-    range_set = RangeSet[T]([ranges[0]])
-    for range in ranges[1:]:
-        if range_set.intersection(range):
-            return True
-        range_set.add(range)
+    stack: list[Range[T]] = []
+    for range in sorted(ranges):
```
ngnpope, line 1058: The tests don't cover non-consecutive overlapping ranges which is important to highlight that this `sorted()` call is necessary.

```python
class TestAnyOverlapping:
    ...

    @pytest.mark.parametrize(
        "ranges_",
        [
            [
                ranges.Range(0, 2),
                ranges.Range(4, 5),  # Added this to ensure the overlapping ranges are not adjacent.
                ranges.Range(1, 3),
            ],
            [
                ranges.Range(
                    0, 2, boundaries=ranges.RangeBoundaries.INCLUSIVE_INCLUSIVE
                ),
                ranges.Range(2, 4),
            ],
        ],
    )
    def test_returns_true_if_and_ranges_overlap_non_consecutive(self, ranges_):
        assert ranges.any_overlapping(ranges_)
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Optimise the `ranges.any_overlapping` function [#184](https://github.com/octoenergy/xocto/pull/184).
```
romainletendart, line 5:
```suggestion
- Improve the performance of the `ranges.any_overlapping` function (with some benchmark showing a >100x speed up) [#184](https://github.com/octoenergy/xocto/pull/184).
```
or any similar phrasing would give a better idea of the actual change.

romainletendart (approved):
Looks good to me, well done! 👏

xocto/ranges.py
```diff
@@ -1054,14 +1054,15 @@ def get_finite_datetime_ranges_from_timestamps(
...
+        if len(stack) == 0:
+            stack.append(range)
+        else:
+            if stack[-1].intersection(range):
+                return True
+            else:
+                stack.append(range)
```
ngnpope, line 1065: We don't need to keep a stack as we're only looking at the last item:

```suggestion
    prev_range: Range[T] | None = None
    for range in sorted(ranges):
        if prev_range is not None and prev_range & range:
            return True
        prev_range = range
```

xocto/ranges.py
Peter554: Yes nice, added

xocto/ranges.py
Peter554: True! That even helped the performance a little more 👍

PR #185, opened 2025-01-13
Add `ranges.any_gaps`

Peter554:
This PR adds a `ranges.any_gaps` utility function. 

This function is very simple - all work is delegated to a `RangeSet`. So arguably this function is not needed at all. I think it's still a useful addition though, since this check is commonly used and this utility function makes it slightly clearer and less likely to make an error.

A benchmark is added, however the existing implementation was found to be pretty optimal. I suggest leaving the benchmark to help guard against regressions (I'm interested to set up codespeed for this, see [here](https://krakentech.slack.com/archives/C01ETJG1RGV/p1736758120170919?thread_ts=1736494580.912409&cid=C01ETJG1RGV)).

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -589,6 +589,17 @@ class TestBreakPeriodsOnTimestamp:
 
 
 class TestAnyOverlapping:
+    def test_does_not_modify_ranges(self):
+        # The implementation of `any_overlapping` relies on sorting.
+        # Let's make sure that the ranges passed in are unchanged.
+        ranges_ = [
+            ranges.Range(1, 2),
+            ranges.Range(0, 1),
+        ]
+        ranges_copy = ranges_.copy()
+        assert not ranges.any_overlapping(ranges_)
+        assert ranges_ == ranges_copy
+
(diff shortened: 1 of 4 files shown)


tests/test_ranges.py
```diff
@@ -589,6 +589,17 @@ def test_timestamps_sorted(self, period):
 
 
 class TestAnyOverlapping:
+    def test_does_not_modify_ranges(self):
```
Peter554, line 592: This should probably have been part of https://github.com/octoenergy/xocto/pull/184/

tests/test_ranges.py
```diff
@@ -624,6 +624,7 @@ def test_does_not_modify_ranges(self):
     )
     def test_returns_true_if_and_ranges_overlap(self, ranges_):
         assert ranges.any_overlapping(ranges_)
+        assert ranges.any_overlapping(reversed(ranges_))
```
Peter554, line 627: This should probably have been part of https://github.com/octoenergy/xocto/pull/184/

leamingrad (approved):
LGTM!

PR #187, opened 2025-01-21
Optimise FiniteDatetimeRange __lt__, intersection and union

Peter554:
This PR optimises the performance of some key `FiniteDatetimeRange` methods: `__lt__`, `intersection` and `union`. This optimisation is achieved by overriding the generic base class implementation with a custom implementation of these methods. These custom implementations exploit the fact that `FiniteDatetimeRange` are always finite and having `INCLUSIVE_EXCLUSIVE` bounds.

This optimisation adds a little complexity to the code (more lines of code is more to maintain). So why should we do this? In Kraken `FiniteDatetimeRange` are used heavily in cost calculations, and these operations can take up a significant amount of the computational time. This optimisation should have give a small but noticeable performance benefit. 

```py
------------------ benchmark 'test_intersection_is_none': 2 tests -----------------
Name (time in ns)                                Mean            Rounds  Iterations
-----------------------------------------------------------------------------------
test_intersection_is_none (NOW)              261.4016 (1.0)      192013          19
test_intersection_is_none (0058_24b4d5c)     960.9678 (3.68)     183218           1
-----------------------------------------------------------------------------------

----------------- benchmark 'test_intersection_is_not_none': 2 tests ----------------
Name (time in us)                                  Mean            Rounds  Iterations
-------------------------------------------------------------------------------------
test_intersection_is_not_none (NOW)              2.8469 (1.0)      190477           1
test_intersection_is_not_none (0058_24b4d5c)     7.1524 (2.51)      79739           1
-------------------------------------------------------------------------------------

----------------- benchmark 'test_sorting': 2 tests ------------------
Name (time in ms)                   Mean            Rounds  Iterations
----------------------------------------------------------------------
test_sorting (NOW)              140.7984 (1.0)           7           1
test_sorting (0058_24b4d5c)     212.3164 (1.51)          5           1
----------------------------------------------------------------------

----------------- benchmark 'test_union_is_none': 2 tests ------------------
Name (time in ns)                         Mean            Rounds  Iterations
----------------------------------------------------------------------------
test_union_is_none (NOW)              264.6992 (1.0)      176461          20
test_union_is_none (0058_24b4d5c)     685.8727 (2.59)      68767          20
----------------------------------------------------------------------------

---------------- benchmark 'test_union_is_not_none': 2 tests -----------------
Name (time in us)                           Mean            Rounds  Iterations
------------------------------------------------------------------------------
test_union_is_not_none (NOW)              2.9004 (1.0)      162154           1
test_union_is_not_none (0058_24b4d5c)     6.4389 (2.22)      53451           1
------------------------------------------------------------------------------
```

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1019,6 +1019,60 @@ class TestFiniteDateRange:
 
 
 class TestFiniteDatetimeRange:
+    @pytest.mark.parametrize(
+        "r1, r2, expected",
+        [
+            [
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 1),
+                    end=datetime.datetime(2000, 1, 3),
+                ),
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 2),
+                    end=datetime.datetime(2000, 1, 4),
+                ),
+                True,
+            ],
+            [
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 2),
+                    end=datetime.datetime(2000, 1, 4),
+                ),
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 1),
+                    end=datetime.datetime(2000, 1, 3),
+                ),
+                False,
+            ],
+            [
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 1),
+                    end=datetime.datetime(2000, 1, 3),
+                ),
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 1),
+                    end=datetime.datetime(2000, 1, 5),
+                ),
+                # False, since only `start` is considered.
+                False,
+            ],
+            [
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 1),
+                    end=datetime.datetime(2000, 1, 3),
+                ),
+                ranges.Range(
+                    start=None,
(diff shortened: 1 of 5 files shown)


tests/test_ranges.py
```diff
@@ -1019,6 +1019,48 @@ def test_finite_range(self):
...
+            [
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 1),
+                    end=datetime.datetime(2000, 1, 3),
+                ),
+                ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2000, 1, 1),
+                    end=datetime.datetime(2000, 1, 5),
+                ),
+                # False, since only `start` is considered.
```
Peter554, line 1056: This is consistent with the base class implementation - `end` is not considered.

xocto/ranges.py
```diff
@@ -837,6 +837,14 @@ def __init__(self, start: datetime.datetime, end: datetime.datetime):
         """
         super().__init__(start, end, boundaries=RangeBoundaries.INCLUSIVE_EXCLUSIVE)
 
+    def __lt__(self, other: Range[datetime.datetime]) -> bool:
+        if isinstance(other, FiniteDatetimeRange):
```
leamingrad, line 841: 🤔 I think this could be optimised further to:

```py3
if other.start is None:
    # We don't need to check anything more if the other range
    # is open-ended
    return False
else:
    return self.start < other.start
```

That should be faster as I think that checking for `None` is quicker than an `isinstance` check.

leamingrad (approved):
LGTM!

PR #189, opened 2025-01-23
Allow union of FiniteDatetimeRange with HalfFiniteDatetimeRange/DatetimeRange

Peter554:
I noticed that it is not possible to take the union between a `FiniteDatetimeRange` and a `HalfFiniteDatetimeRange` (or a general `DatetimeRange`). The pre-existing implementation tries to force the result of `FiniteDatetimeRange.union` to also be finite, at which point the implementation errors since the union of a `FiniteDatetimeRange` and a `HalfFiniteDatetimeRange` will not in general be finite. 

This seems wrong - it should be possible to take the union of a `FiniteDatetimeRange` and a `HalfFiniteDatetimeRange`.

Maybe there has been some confusion with intersection here. The intersection of an `FiniteDatetimeRange` with a `HalfFiniteDatetimeRange` will always be finite, but the union will not be.

We could (likely should) make equivalent changes to `FiniteDateRange`, however I'd leave that for a separate PR.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1124,7 +1124,6 @@ class TestFiniteDatetimeRange:
                 )
             )
 
-        @pytest.mark.xfail
         @pytest.mark.parametrize(
             "range_, other, expected_union",
             [
@@ -1162,7 +1161,6 @@ class TestFiniteDatetimeRange:
         def test_union_with_half_finite_range(self, range_, other, expected_union):
             assert range_ | other == other | range_ == expected_union
 
-        @pytest.mark.xfail
         def test_union_with_continuum(self):
             range = ranges.FiniteDatetimeRange(
                 datetime.datetime(2020, 1, 1),
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1084,6 +1084,7 @@ class TestFiniteDatetimeRange:
                 end=datetime.datetime(2000, 1, 3),
             )
 
+            assert isinstance(range | other, ranges.FiniteDatetimeRange)
             assert (
                 range | other
                 == other | range
@@ -1115,6 +1116,7 @@ class TestFiniteDatetimeRange:
                 end=datetime.datetime(2000, 1, 4),
             )
 
+            assert isinstance(range | other, ranges.FiniteDatetimeRange)
             assert (
                 range | other
                 == other | range
@@ -1159,6 +1161,7 @@ class TestFiniteDatetimeRange:
             ],
         )
         def test_union_with_half_finite_range(self, range_, other, expected_union):
+            assert ranges._is_half_finite_datetime_range(range_ | other)
             assert range_ | other == other | range_ == expected_union
 
         def test_union_with_continuum(self):
@@ -1168,6 +1171,7 @@ class TestFiniteDatetimeRange:
             )
             other = ranges.DatetimeRange.continuum()
 
+            assert ranges._is_datetime_range(range | other)
             assert range | other == other | range == ranges.DatetimeRange.continuum()
 
     class TestIntersection:
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1124,6 +1124,44 @@ class TestFiniteDatetimeRange:
                 )
             )
 
+        @pytest.mark.xfail
+        @pytest.mark.parametrize(
+            "range_, other, expected_union",
+            [
+                [
+                    ranges.FiniteDatetimeRange(
+                        datetime.datetime(2020, 1, 1),
+                        datetime.datetime(2023, 1, 1),
+                    ),
+                    ranges.HalfFiniteDatetimeRange(
+                        datetime.datetime(2021, 1, 1),
+                        None,
+                    ),
+                    ranges.HalfFiniteDatetimeRange(
+                        start=datetime.datetime(2020, 1, 1),
+                        end=None,
+                    ),
+                ],
+                [
+                    ranges.FiniteDatetimeRange(
+                        datetime.datetime(2020, 1, 1),
+                        datetime.datetime(2023, 1, 1),
+                    ),
+                    ranges.HalfFiniteDatetimeRange(
+                        datetime.datetime(2021, 1, 1),
+                        datetime.datetime(2022, 1, 1),
+                    ),
+                    # The resulting range is finite, but it's still an instance of `HalfFiniteDatetimeRange`.
+                    ranges.HalfFiniteDatetimeRange(
+                        start=datetime.datetime(2020, 1, 1),
+                        end=datetime.datetime(2023, 1, 1),
+                    ),
+                ],
+            ],
+        )
+        def test_union_with_half_finite_range(self, range_, other, expected_union):
+            assert range_ | other == other | range_ == expected_union
+
     class TestIntersection:
         def test_intersection_of_touching_ranges(self):
             range = ranges.FiniteDatetimeRange(
@@ -1165,6 +1203,25 @@ class TestFiniteDatetimeRange:
                 == ranges.FiniteDatetimeRange(
                     start=datetime.datetime(2000, 1, 2),
                     end=datetime.datetime(2000, 1, 3),
+                )
+            )
+
+        def test_intersection_with_half_finite_range(self):
+            range = ranges.FiniteDatetimeRange(
+                datetime.datetime(2020, 1, 1),
+                datetime.datetime(2022, 1, 1),
+            )
+            other = ranges.HalfFiniteDatetimeRange(
+                datetime.datetime(2021, 1, 1),
+                None,
+            )
+
+            assert (
+                range & other
+                == other & range
+                == ranges.FiniteDatetimeRange(
+                    start=datetime.datetime(2021, 1, 1),
+                    end=datetime.datetime(2022, 1, 1),
                 )
             )
 
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1162,6 +1162,16 @@ class TestFiniteDatetimeRange:
         def test_union_with_half_finite_range(self, range_, other, expected_union):
             assert range_ | other == other | range_ == expected_union
 
+        @pytest.mark.xfail
+        def test_union_with_continuum(self):
+            range = ranges.FiniteDatetimeRange(
+                datetime.datetime(2020, 1, 1),
+                datetime.datetime(2022, 1, 1),
+            )
+            other = ranges.DatetimeRange.continuum()
+
+            assert range | other == other | range == ranges.DatetimeRange.continuum()
+
     class TestIntersection:
         def test_intersection_of_touching_ranges(self):
             range = ranges.FiniteDatetimeRange(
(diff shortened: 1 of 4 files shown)


tests/test_ranges.py
```diff
@@ -1124,6 +1124,26 @@ def test_union_of_overlapping_ranges(self):
                 )
             )
 
+        @pytest.mark.xfail
```
Peter554: xfail here since in this commit the test will fail - the xfail is removed later in the PR.

tests/test_ranges.py
```diff
@@ -1144,6 +1144,16 @@ def test_union_with_half_finite_range(self):
...
```
Peter554: xfail here since in this commit the test will fail - the xfail is removed later in the PR.

xocto/ranges.py
```diff
@@ -1151,3 +1165,19 @@ def iterate_over_months(
...
+    if value.start is not None and not isinstance(value.start, datetime.datetime):
+        return False
+    if value.end is not None and not isinstance(value.end, datetime.datetime):
+        return False
+    return True
+
+
+# Subscripted generics may not be used with isinstance directly.
+# TODO: A TypeGuard would be nicer, once we drop Python 3.9.
+def _is_half_finite_datetime_range(value: Any) -> bool:
```
Peter554, line 1199: I moved this function from `xocto/fields/postgres/ranges` to here, since I need the same function here and this seems like the better home for it. It means we're calling a private function of `xocto/ranges` in `xocto/fields/postgres/ranges`, but I'm not aware of a better way to do this. I don't want to make this function public (`def is_half_finite_datetime_range`), since then it would also be part of the external public API. Ideally I'd want some kind of "public within this package" access, but python doesn't have anything like that... I think this is probably fine as it is 🤷

xocto/fields/postgres/ranges.py
```diff
@@ -233,14 +233,3 @@ def value_to_string(self, obj: models.Model) -> Optional[str]:
                 "end": base_field.value_to_string(end),
             }
         )
-
-    def _is_half_finite_datetime_range(self, value: Any) -> bool:
```
Peter554, line 237: ♻️ Moved to `xocto/ranges` -> https://github.com/octoenergy/xocto/pull/189#discussion_r1933268951

xocto/ranges.py
```diff
@@ -891,10 +903,12 @@ def union(self, other: Range[datetime.datetime]) -> Optional["FiniteDatetimeRang
         if base_union is None:
             return None
 
-        assert base_union.boundaries == RangeBoundaries.INCLUSIVE_EXCLUSIVE
-        assert base_union.start is not None
-        assert base_union.end is not None
-        return FiniteDatetimeRange(base_union.start, base_union.end)
+        if _is_half_finite_datetime_range(other):
+            return HalfFiniteDatetimeRange(
```
Peter554, line 907: I'm deliberately calling the constructor here rather than using `cast` (we use `cast` in some other places in `xocto/ranges`) - see https://github.com/octoenergy/xocto/issues/193

xocto/ranges.py
```diff
@@ -1151,3 +1182,19 @@ def iterate_over_months(
...
+
+
+# Subscripted generics may not be used with isinstance directly.
+# TODO: A TypeGuard would be nicer, once we drop Python 3.9.
+def _is_datetime_range(value: Any) -> bool:
+    if value.start is not None and not isinstance(value.start, datetime.datetime):
+        return False
+    if value.end is not None and not isinstance(value.end, datetime.datetime):
+        return False
+    return True
```
yuwash, line 1194: That’s a lot of `not`s and `False`s, I might find it easier to understand by simply putting everything together as a positive expression:

```suggestion
    return (
        (value.start is None or isinstance(value.start, datetime.datetime))
        and (value.end is None or isinstance(value.end, datetime.datetime))
    )
```

xocto/ranges.py
```diff
@@ -891,16 +903,35 @@ def union(self, other: Range[datetime.datetime]) -> Optional["FiniteDatetimeRang
...
-        assert base_union.boundaries == RangeBoundaries.INCLUSIVE_EXCLUSIVE
-        assert base_union.start is not None
-        assert base_union.end is not None
-        return FiniteDatetimeRange(base_union.start, base_union.end)
+        if _is_half_finite_datetime_range(other):
+            return HalfFiniteDatetimeRange(
+                cast(datetime.datetime, base_union.start), base_union.end
+            )
+        else:
+            return base_union
```
yuwash, line 911: The implementation seems to be a bit inconsistent with the existing one from the base class (or from `isinstance(other, FiniteDatetimeRange)` which allegedly is equivalent to the base class implementation, just more performant): for the resulting range to be half-finite, the `other` range needs to be ending later than the `self` range. Otherwise it should result in a finite range. The base class implementation does that in a somewhat obscure bounds juggling. https://github.com/octoenergy/xocto/blob/956e83b1fda6238483ad3b3af1df97ee4fb9d450/xocto/ranges.py#L404-L406

Maybe what we needed is just to remove the `assert base_union.end is not None` line above? Or what was the failure that you were experiencing before the change? Of course we could also make a specialized efficient implementation analogous to the `isinstance(other, FiniteDatetimeRange)` case, but that would only make sense if put before the `super().union(other)` call to actually save time.

tests/test_ranges.py
```diff
@@ -1124,6 +1126,36 @@ def test_union_of_overlapping_ranges(self):
                 )
             )
 
+        def test_union_with_half_finite_range(self):
```
yuwash, line 1129: Great to have some tests for this which I wonder it was already somehow implemented, but it would be good to also have a test where
* we have a `other.end <= range.end` (`==` is also an interesting edge case) -> should result in a finite range
* we have a non-open end in the half-finite one

xocto/ranges.py
Peter554: Sure, that's probably a little more readable - updated.

tests/test_ranges.py
Peter554: > which I wonder it was already somehow implemented

The tests failed (hence the `xfail` in the first commit) - it's only after my changes that they passed.

> we have a other.end <= range.end (== is also an interesting edge case) -> **should result in a finite range**

I'm not quite sure I follow. I believe that `FiniteDatetimeRange | HalfFiniteDatetimeRange` should _always_ be of type `HalfFiniteDatetimeRange`.

Are you perhaps mixing up a "dt range that is finite" with "being an actual instance of `FiniteDatetimeRange`"? A `HalfFiniteDatetimeRange` can always be finite if we set non-`None` boundaries, it just doesn't make any guarantees to be finite to the type system.

I could update the typing of `FiniteDatetimeRange.union` to be

```py
@overload
def union(
    self, other: HalfFiniteDatetimeRange
) -> Optional[FiniteDatetimeRange | HalfFiniteDatetimeRange]: ...
```

and then in the implementation I'd need to check if the resulting range is in fact finite and if so return the `FiniteDatetimeRange` class and else the `HalfFiniteDatetimeRange`. However, I thinkit's easier for everyone to just always return `HalfFiniteDatetimeRange`.
* It's simpler in the implementation (easier to read and more performant - one less `if`).
* It's also simpler for the caller. The return type `Optional[FiniteDatetimeRange | HalfFiniteDatetimeRange]` is not very useful - if the caller really wanted to use the `FiniteDatetimeRange` then the caller will have to use `isinstance` checks to narrow down the type further themselves.

xocto/ranges.py
Peter554: ♻️ -> https://github.com/octoenergy/xocto/pull/189#discussion_r1935203040

Peter554: > Or what was the failure that you were experiencing before the change?

The failure is in the typing more than in the implementation. On main currently we have

```py
def union(self, other: Range[datetime.datetime]) -> Optional["FiniteDatetimeRange"]:
```

Says that the union of a `FiniteDatetimeRange` with a general `DatetimeRange` is of type `FiniteDatetimeRange`, which simply isn't true (it is true, but only because we error if the resulting union is not finite).

So the main change in this PR is to update typing to be more expressive:

```py
@overload
def union(self, other: FiniteDatetimeRange) -> Optional[FiniteDatetimeRange]: ...

@overload
def union(
    self, other: HalfFiniteDatetimeRange
) -> Optional[HalfFiniteDatetimeRange]: ...

@overload
def union(self, other: DatetimeRange) -> Optional[DatetimeRange]: ...
```

The actual implementation changes are then just following along naturally.

tests/test_ranges.py
Peter554: > we have a other.end <= range.end (== is also an interesting edge case) -> should result in a finite range

I added a test case now for completeness, but I'm asserting that the result is an instance of `HalfFiniteDatetimeRange` rather than `FiniteDatetimeRange` since this is the result that I think we want.

xocto/ranges.py
yuwash: Probably what can be done is to make a “private package” and make the function itself public but that would involve a major restructuring so not actually suggesting now

tests/test_ranges.py
yuwash: Aha, I was previously thinking the use of `RangeBoundaries.INCLUSIVE_EXCLUSIVE` were the main characteristic of `HalfFiniteRange` whilst `FiniteRange` would be inclusive both sides, but apparently for the latter you can use any boundaries you want. Now I understand that indeed `FiniteRange` is _almost_ a subset of `HalfFiniteRange` apart from there being no such relationship in the actual implementation (and the classes not being duck types).

Another thing is though also the boundary constraint that’s only present in `HalfFiniteRange` (that what I was thinking to be the main characteristic). Thus it might not be always possible to guarantee

https://github.com/octoenergy/xocto/blob/956e83b1fda6238483ad3b3af1df97ee4fb9d450/xocto/ranges.py#L876-L879

in case `self.boundaries` has an exclusive start or an inclusive end and one of those bounds would be larger than the corresponding bound of `other`. It’s good that you aren’t casting it to make it look like a `HalfFiniteRange` when it’s not, but by using the constructor and ignoring the `base_union.boundaries`, it’s possibly
* wrongly making `base_union.start` inclusive when it was exclusive and/or
* losing the inclusivity of `base_union.end`

Peter554: Hmm I'm still not sure I quite agree. 
* Inside of `if _is_half_finite_datetime_range(other)` we can be sure that both `self` and `other` have boundaries INCLUSIVE_EXCLUSIVE. Since `self` is `FiniteDatetimeRange` and `other` is `HalfFiniteDatetimeRange. 
* Since both `self` and `other` have boundaries INCLUSIVE_EXCLUSIVE, we can be sure that their union also has these boundaries.

I've updated the code now with some comments to hopefully make this clearer. What do you think?

yuwash: Ah true, `FiniteDatetimeRange` has the same boundaries as `HalfFiniteRange` while `FiniteRange` doesn’t. I was blindly thinking `FiniteDatetimeRange` is just `FiniteRange[datetime.datetime]` analogous to `HalfFiniteDatetimeRange` :see_no_evil:
_Your added logic_ is all good! (while the whole thing appears to have been a bit inconsistent)

Also :100: helpful comment you added!

PR #190, opened 2025-01-24
Fix computation of FiniteDatetimeRange.days

Peter554:


Peter554:
Closing this for now - on reflection I think it's not so simple. See https://github.com/octoenergy/xocto/issues/192

Issue #192, opened 2025-01-24
Care around datetime ranges with mixed tzifo / fix `FiniteDatetimeRange.days`


Peter554:
## Problem summary
* It's possible to construct `FiniteDatetimeRange`s with mixed tzinfo (where the start and end have different tzinfo). 
* This makes the `FiniteDatetimeRange.days` property confusing and error prone (Kraken: important for computation of daily standing charges). For example, is the below range 30 or 31 days?

```py
RANGE = ranges.FiniteDatetimeRange(
    # This is also 2024-03-01T00:00:00 in TZ_UTC
    datetime.datetime(2024, 3, 1, tzinfo=TZ_LONDON),

    # This is 2024-04-01T00:00:00 in TZ_LONDON
    datetime.datetime(2024, 3, 31, hour=23, tzinfo=TZ_UTC),
)
```

* We're unlikely to construct such ranges directly, but it's possible to create them accidentally via intersection/union operations. For example, if ranges coming from database records are typically going to be UTC, but other ranges are often created in localtime. 
* This issue was brought to the spotlight by some failing tests when trying to upgrade xocto in Kraken from v7.1.0 -> v7.1.1, but it's actually symptom of a  pre-existing issue.

## Problem in detail

See https://github.com/octoenergy/xocto/pull/191/files, which demonstrates:
*  For a `FiniteDatetimeRange` with mixed tzinfo the `.days` property is ambiguous. 
* `(r1 & r2).days` can be different from `(r2 & r1).days`.
* `(r1 | r2).days` can be different from `r1.days` even when `r2 == r1`.
* This demo is branched from v7.1.0, demonstrating that this issue existed before v7.1.1.

## Proposed solution
* [Breaking change] Remove the `FiniteDatetimeRange.days` property.
* Instead, require callers to convert to a date range and compute days that way e.g. `dt_range.localize(tz=...).as_date_range().days`.

benthorner:
## 🐮 Thoughts about `.days`

This method is misleading for a range at time granularity: it should return a `Decimal` or be named e.g. `whole_days`.

## Thoughts about ambiguity

> This makes the FiniteDatetimeRange.days property ambiguous

Could you elaborate on this point? I can see some examples in the linked PR e.g.

```python
# This range has inconsistent tzinfo.
# It's unclear how many days this represents.
range_utc = ranges.FiniteDatetimeRange(
    datetime.datetime(2024, 3, 1, tzinfo=TZ_LONDON),
[...]

PR #194, opened 2025-01-29
Disallow datetime ranges with mixed tzinfo

Peter554:
Context: https://github.com/octoenergy/xocto/issues/192

---

Update: I've tried this in Kraken but the number of errors was rather daunting. I think it's unlikely we find time to prioritise this change right now, so I'm proposing https://github.com/octoenergy/xocto/pull/195 as an alternative/first-step to unblock things.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -1019,6 +1019,15 @@ class TestFiniteDateRange:
 
 
 class TestFiniteDatetimeRange:
+    def test_cannot_construct_with_inconsistent_tzinfo(self):
+        with pytest.raises(ranges.InconsistentTzInfo):
+            tz_london = zoneinfo.ZoneInfo("Europe/London")
+            tz_berlin = zoneinfo.ZoneInfo("Europe/Berlin")
+            ranges.FiniteDatetimeRange(
+                datetime.datetime(2020, 1, 1, tzinfo=tz_london),
+                datetime.datetime(2021, 1, 1, tzinfo=tz_berlin),
+            )
+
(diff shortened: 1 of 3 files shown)


tests/test_ranges.py
```diff
@@ -1243,19 +1252,6 @@ def test_returns_date_range(self):
                 datetime.date(2020, 1, 9),
             )
 
-        def test_errors_if_different_timezones(self):
-            dt_range = ranges.FiniteDatetimeRange(
-                datetime.datetime(2020, 1, 1, tzinfo=zoneinfo.ZoneInfo("Asia/Dubai")),
-                datetime.datetime(
-                    2020, 1, 10, tzinfo=zoneinfo.ZoneInfo("Australia/Sydney")
-                ),
-            )
```
Peter554, line 1252: Removing this test case since it's no longer possible to construct such a range!

tests/test_ranges.py
```diff
@@ -1019,6 +1019,15 @@ def test_finite_range(self):
 
 
 class TestFiniteDatetimeRange:
+    def test_cannot_construct_with_inconsistent_tzinfo(self):
```
Peter554, line 1026: It would be nice to have similar logic for `DatetimeRange` and `HalfFiniteDatetimeRange`, but those two are currently not independent classes but just type aliases. So there is nowhere to insert the logic 🤔 Is there any reason not to make `DatetimeRange` and `HalfFiniteDatetimeRange` actual classes (if we decide to do this I might save that for a separate PR, or do it as a preparation PR)?

PR #195, opened 2025-01-30
Update `FiniteDatetimeRange.days` to require passing a TZ

Peter554:
It's possible to create mixed TZ FiniteDatetimeRange's. It probably shouldn't be, but it currently is and this is unlikely to change too soon. See https://github.com/octoenergy/xocto/issues/192

This makes the `FiniteDatetimeRange.days` property ambiguous. For example, should the below range be 31 or 30 days?

```py
RANGE = ranges.FiniteDatetimeRange(
    # This is also 2024-03-01T00:00:00 in TZ_UTC
    datetime.datetime(2024, 3, 1, tzinfo=TZ_LONDON),

    # This is 2024-04-01T00:00:00 in TZ_LONDON
    datetime.datetime(2024, 3, 31, hour=23, tzinfo=TZ_UTC),
)
```

Getting this wrong can lead to bugs in e.g. computation of per-day standing costs.

This PR updates `FiniteDatetimeRange.days` to require passing a TZ, to remove this ambiguity. This is a bit more verbose, but it's much more explicit and less error prone.

### DIFF
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -860,7 +860,7 @@ class FiniteDatetimeRange(FiniteRange[datetime.datetime]):
             if left.end <= right.start:
                 return None
             else:
-                return FiniteDatetimeRange(right.start, min(left.end, right.end))
+                return FiniteDatetimeRange(right.start, min(right.end, left.end))
 
         base_intersection = super().intersection(other)
         if base_intersection is None:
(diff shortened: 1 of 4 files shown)


xocto/ranges.py
```diff
@@ -901,12 +901,12 @@ def __and__(
...
         return self.intersection(other)
 
-    @property
-    def days(self) -> int:
+    def days(self, tz: datetime.tzinfo) -> int:
         """
         Return the number of days between the start and end of the range.
         """
-        return (self.end - self.start).days
+        range_ = self.localize(tz)
```
benthorner, line 908: ❓ How about disallowing differing zones, just for this property?

Requiring `tz` means we're going to need to update all call sites anyway, so it's easy enough to do `period.localize(tz).days` instead of `period.days(tz)`. That keeps this property simple and makes the caller a little more explicit with its intent.

PR #196, opened 2025-01-30
Reverse order of `min` in `FiniteDatetimeRange.intersection`

Peter554:
In https://github.com/octoenergy/xocto/pull/187 the implementation of `FiniteDatetimeRange.intersection` was optimised. This introduced a very subtle change in behaviour. Previously when `left.end` was equal to `right.end` then the `right.end` was favoured. In https://github.com/octoenergy/xocto/pull/187 the implementation was (unintentionally) changed to favour `left.end`.

This change reverses that, to again favour `right.end`.

This really shouldn't matter, but it can be significant when the TZ of left and right are different. Eventually we'd like to prevent this kind of mixed TZ behaviour (see https://github.com/octoenergy/xocto/issues/192), but in the short term it seems reasonable to make this tiny change to make things consistent again with how they were before and in order to unblock xocto releases / install in Kraken.

### DIFF
--- a/tests/test_ranges.py
+++ b/tests/test_ranges.py
@@ -13,6 +13,10 @@ from hypothesis.strategies import composite, integers, none, one_of, sampled_fro
 from xocto import localtime, ranges
 
 
+TZ_LONDON = zoneinfo.ZoneInfo("Europe/London")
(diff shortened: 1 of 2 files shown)


leamingrad (approved):
LGTM, but could we add a test to cover this situation (since it is subtle behaviour)?

PR #198, opened 2025-01-31
Refer to ddtrace.tracer from a non deprecated location

delfick:
Prior to this the recent upgrade of ddtrace deprecated accessing tracer from ddtrace.

It is instead exported from ddtrace.trace now

### DIFF
--- a/xocto/tracing.py
+++ b/xocto/tracing.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-import ddtrace
+import ddtrace.trace
 
 
-tracer = ddtrace.tracer
+tracer = ddtrace.trace.tracer
 wrap = tracer.wrap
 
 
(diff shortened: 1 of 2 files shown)


xocto/tracing.py
```diff
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
-import ddtrace
+import ddtrace.trace
 
 
-tracer = ddtrace.tracer
+tracer = ddtrace.trace.tracer
```
Peter554, line 6: Thanks for looking at this. 

Is this going to work with Kraken? Kraken is using ddtrace version 2.14.4 https://github.com/octoenergy/kraken-core/blob/b7ca6037788bf6101ef655ba508f4d706fe2df2f/requirements/prod.txt#L142 and in 2.14.4 this code will raise `AttributeError: module 'ddtrace.trace' has no attribute 'tracer'`.

Peter554: In the meantime https://github.com/octoenergy/xocto/pull/199

delfick:
we instead chose to remove ddtrace from xocto altogether https://github.com/octoenergy/xocto/pull/201

PR #200, opened 2025-02-02
Intervaltree

Peter554:
TODO:
* Decide about "insert same interval with equal data"
* Describe motivation 
* Describe differences from intervaltree library
* Describe why not using ranges
* Docs
* Changelog


Peter554:
Closing since I'm not actively working on it.

PR #205, opened 2025-02-14
Remove `FiniteDatetimeRange.days`

Peter554:
See issue https://github.com/octoenergy/xocto/issues/192 for context.


Peter554:
I'm going to close this one for now. We'd like to remove this property, but we need to burn down the usages in Kraken first.

PR #206, opened 2025-02-14
Add `localtime.get_local_timezone()`

Peter554:
This PR adds `localtime.get_local_timezone()`. This avoids having to remember to import `django.utils.timezone` and call `timezone.get_current_timezone()`. It's only a small improvement, but it helps a little IMO.


agapius (approved):
lgtm!

PR #209, opened 2025-03-03
Benchmark improvements

Peter554:
This PR:

* Makes sure that we're running benchmarks in CI (just once), so that we can be sure those test cases still run properly. This is objectively a good idea.
* Moves benchmarks to a separate top level folder and updates pytest options so that they are discovered properly. This one is more subjective, but I think it's nice to separate benchmarks in this way.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -217,7 +217,9 @@ filterwarnings = [
 DJANGO_SETTINGS_MODULE = "tests.settings"
 
 # Test modules must have be named this way.
-python_files = "test_*.py"
+python_files = ["test_*.py", "benchmark_*.py"]
+python_functions = ["test_*", "benchmark_*"]
+python_classes = ["Test*", "Benchmark*"]
 
 # Default options when pytest is run:
 #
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -217,7 +217,9 @@ filterwarnings = [
 DJANGO_SETTINGS_MODULE = "tests.settings"
 
 # Test modules must have be named this way.
-python_files = "test_*.py"
+python_files = ["test_*.py", "benchmark_*.py"]
+python_functions = ["test_*", "benchmark_*"]
+python_classes = ["Test*", "Benchmark*"]
```
Peter554, line 222: You can see that the benchmarks are being run correctly e.g. here https://app.circleci.com/pipelines/github/octoenergy/xocto/802/workflows/32934bd1-2131-43ef-a135-474ea164ee21/jobs/781/parallel-runs/0/steps/0-110?invite=true#step-110-3189_79

agapius (approved):
- We do want to ensure benchmark tests don’t break in CI.
- I see that we already test the ranges module in `tests/test_ranges.py`. The functions in `tests/benchmarks/test_ranges.py` are exclusively there to benchmark. So it also makes sense to move them to a separate top level folder.

🚢 

PS: Reading the code, I was a bit confused by the assertions in the benchmarks: I thought here we combined testing and benchmarking. But I later discovered that we keep the assert just as some correctness guarantee for the benchmark (theoretically we could remove the assert).

PR #210, opened 2025-03-03
Start storing benchmarks data

Peter554:
As part of performance work for costing, we added some benchmarks to xocto (for the ranges module). These benchmarks were useful to measure our improvements locally. However they are not being run in CI and we're not keeping a record of the results over time. So they only have limited use right now.

This PR proposes a very lightweight solution to start running benchmarks in CI and storing the results.

* A GitHub workflow to run benchmarks on every push to `main` and store the results as a CSV in a dedicated `benchmarks-data` branch.
* A GitHub workflow to collect benchmark result CSVs into a single, consolidated CSV living on the `benchmarks-data` branch.

Given the above workflows, it would make sense to make `benchmarks-data` a protected branch.

For the purpose of this PR we don't do anything with the benchmark results CSV living on the `benchmarks-data` branch. In the future we might like to do things like:
* Create absolute or relative thresholds for the benchmarks (e.g. defined in a config file) and then alert interested parties if those thresholds are crossed (e.g. slack channel with benchmark alerts).
* Create another workflow to run benchmarks on PRs and use the results from main to compare. Was there a regression/improvement? We can automatically leave a comment on the PR to flag any issues or even fail the build.
* Analyse the data e.g. change point detection.

Basically, once we start collecting some data there are lots of options for how to take this further and make it more useful.

I've already created the `benchmarks-data` branch here https://github.com/octoenergy/xocto/tree/benchmarks-data.

Why this custom approach?. Why not use a 3rd party service? True. I have been looking into https://codspeed.io/ and https://bencher.dev/ as alternatives. There are a few issues here:
* Main issue: They cost quite a bit of money (actually, for open source projects they are free, but I'm only using xocto as a testing ground. I am really hoping eventually to take this approach and apply it to benchmarks in Kraken).
* Neither platform seems to be quite ideal. I've tested both out and they seem pretty good, but I'm not sure we want to commit to using either.

If we do decide we want to use one of these third party solutions later that's also fine - we can quickly delete this custom solution. I.e. we're not backing ourselves into a corner here at all.

The approach taken here is somewhat inspired by https://github.com/benchmark-action/github-action-benchmark.

For general information about benchmarking in python see https://pytest-benchmark.readthedocs.io/en/latest/.

I've already tested this approach out on my fork https://github.com/Peter554/xocto. See:
* The workflow that runs the benchmarks https://github.com/Peter554/xocto/actions/runs/13653420832
* The workflow that collects the benchmarks https://github.com/Peter554/xocto/actions/runs/13653453022
* The benchmarks data file that is generated on the `benchmarks-data` branch https://github.com/Peter554/xocto/blob/benchmarks-data/benchmarks.csv

### DIFF
--- a/.github/workflows/collect_benchmarks.yml
+++ b/.github/workflows/collect_benchmarks.yml
@@ -1,0 +1,45 @@
+name: Collect benchmarks
+
+on:
+  workflow_run:
+    workflows: [ "Run benchmarks main" ]
+    types:
+      - completed
+
+# `concurrency` used to ensure that only one instance of this workflow runs at one time.
+# This is important since concurrent workflows would operate on the same files.
+concurrency:
+  group: ${{ github.workflow }}
+
+jobs:
+  collect_benchmarks:
+    name: Collect benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout benchmarks-data branch
+        uses: actions/checkout@v4
+        with:
+          ref: benchmarks-data
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      - name: Collect benchmarks
+        run: uv run scripts/collect_benchmarks.py -i 'uncollected/*.csv' -o benchmarks.csv
+
+      - name: Remove now collected benchmarks
+        run: rm uncollected/*.csv
+
+      - name: Add and commit
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          
+          git add --all
+          git commit -m "Collect benchmarks"
+          
+          # Retry below incase another workflow pushed to the benchmarks-data branch since we checked it out.
+          # Unlikely but technically possible...
+          # There won't be a merge conflict since this workflow is the only workflow touching benchmarks.csv
+          # and this workflow uses `concurrency`.
+          scripts/push_with_retry.sh 5
--- a/.github/workflows/run_benchmarks_main.yml
+++ b/.github/workflows/run_benchmarks_main.yml
@@ -1,0 +1,76 @@
+name: Run benchmarks main
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  run_benchmarks:
+    name: Run benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Setup python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Install dependencies
+        run: make install
+
+      - name: Run benchmarks
+        run: pytest --benchmark-only --benchmark-json benchmarks.json
+
+      - name: Upload artifact - benchmarks JSON
+        uses: actions/upload-artifact@v4
+        with:
+          name: benchmarks_json
+          path: benchmarks.json
+          # Short retention. Only needed short term to share with the `extract_benchmarks` job.
+          retention-days: 1
+
+  extract_benchmarks:
+    name: Extract benchmarks
+    runs-on: ubuntu-latest
+    needs: [ run_benchmarks ]
+    steps:
+      - name: Checkout benchmarks-data branch
+        uses: actions/checkout@v4
+        with:
+          ref: benchmarks-data
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      - name: Download artifact - benchmarks JSON
+        uses: actions/download-artifact@v4
+        with:
+          name: benchmarks_json
+
+      - name: Extract benchmarks
+        run: |
+          uv run scripts/extract_benchmarks.py \
+            -i benchmarks.json \
+            -o "uncollected/$(uuidgen).csv" \
+            -t ubuntu-latest
+
+      - name: Remove benchmarks JSON
+        run: rm benchmarks.json
+
+      - name: Add and commit benchmarks data
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          
+          git add --all
+          git commit -m "Add uncollected benchmarks"
+
+          # Retry below incase another workflow pushed to the benchmarks-data branch since we checked it out.
+          # Unlikely but technically possible...
+          # There won't be a merge conflict since we used a UUID as the file name.
+          scripts/push_with_retry.sh 5


scripts/extract_benchmarks.py
```diff
@@ -0,0 +1,67 @@
+"""
+Extracts benchmarks data from the JSON output of pytest-benchmark
```
Peter554, line 2: Why not just keep the entire JSON? The entire JSON is very big. It contains data for every single iteration (can be 1000's per benchmark). We only want/need the summary statistics.

Peter554, line 2: JSON from https://pytest-benchmark.readthedocs.io/en/latest/usage.html#commandline-options with `--benchmark-json`

scripts/extract_benchmarks.py
```diff
@@ -0,0 +1,67 @@
...
+
+def _parse_datetime(s: str) -> datetime.datetime:
+    return datetime.datetime.fromisoformat(s).astimezone(pytz.UTC)
+
+
+def main():
+    arg_parser = argparse.ArgumentParser()
+    arg_parser.add_argument("--input", "-i", required=True)
+    arg_parser.add_argument("--output", "-o", required=True)
+    arg_parser.add_argument("--testbed", "-t", required=True)
```
Peter554, line 24: "testbed" is a term I've taken from https://bencher.dev/. It means the machine the benchmarks were run on (so we should only compare benchmark results that have the same testbed).

.github/workflows/run_benchmarks_main.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Checkout benchmarks-data branch
+        uses: actions/checkout@v4
+        with:
+          ref: benchmarks-data
+          path: benchmarks-data
+
+      - name: Add and commit benchmarks data
+        run: |
+          uuid=$(uuidgen)
+          cp ../main/benchmarks.csv "uncollected/$uuid.csv"
```
Peter554, line 45: UUID is used as the filename here. The filename isn't important (we'll collect the CVSs using a glob), so long as it is unique.

.github/workflows/collect_benchmarks.yml
```diff
@@ -0,0 +1,42 @@
+name: Collect benchmarks
+
+on:
+  workflow_run:
+    workflows: ["Run benchmarks main"]
+    types:
+      - completed
+
+concurrency:
+  group: ${{ github.workflow }}
```
Peter554, line 10: This bit about concurrency ensures that only one of these "Collect benchmarks" workflows runs at the same time. This is important since concurrent workflows would operate on the same files (could result in duplicated or lost data).

pyproject.toml
```diff
@@ -48,6 +48,7 @@ dev = [
   "mypy-boto3-s3==1.34.120",
   "mypy==1.10.0",
   "numpy==1.22.2",
+  "polars>=1.24.0",
```
Peter554, line 51: Will be more useful later on in further work when we want to do analysis of the stored benchmarks data. I'm interested to explore using polars in more places (maybe also in Kraken) and this seems a safe place to play with it (it's only added as a dev dependency).

.github/workflows/run_benchmarks_main.yml
```diff
@@ -0,0 +1,51 @@
...
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add "uncollected/$uuid.csv"
+          git commit -m "Add uncollected benchmarks $uuid.csv"
+          git push
```
Peter554, line 50: TODO Thinking about it more I think this approach is flawed - the git push will fail if any of the other workflows pushed in the meantime. Need to think about this.

.github/workflows/collect_benchmarks.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Remove now collected benchmarks
+        run: rm uncollected/*
+
+      - name: Add and commit
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add --all
+          git commit -m "Collect benchmarks"
+          git push
```
Peter554, line 42: ♻️ https://github.com/octoenergy/xocto/pull/210#discussion_r1977823565

.github/workflows/collect_benchmarks.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  collect_benchmarks:
+    name: Collect benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
```
goldenrati0, line 20: ❓ Why are we running on 3.9 only? Projects python version requirement is `>=3.9`, should we use a matrix and run on multiple python versions?
Downside is CI will be even slower.

.github/workflows/run_benchmarks_main.yml
goldenrati0: Can we export this data to an external data store? Like AWS S3 bucket maybe? Using git for this does not seem ideal because we are not tracking versions for the benchmark files, are we?

scripts/extract_benchmarks.py
goldenrati0: If this is running as a part of Github Workflow, how are we going to make sure that we keep using the same set of machines? AFAIK, this is only possible if we use [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow).

.github/workflows/run_benchmarks_main.yml
Peter554: Okay I've addressed this by implementing a "push with retry".
* Try to push, if it fails do `git pull --rebase`. Loop up to 5 times.
* We know there will never be a merge conflict - see my comments in the job where we are pushing.
* This is really an edge case anyway - we'd only run into issues if there was really high concurrency (very unlikely).

scripts/extract_benchmarks.py
Peter554: I believe GitHub-hosted runners are standardised - see [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). I've used GitHub-hosted runners in this way for another project I'm working on (https://github.com/seddonym/grimp/) and so far had good enough results. 

GitHub may occasionally change the specification for these machines, but so long as this is rare then this is good enough I think (we can just purge old benchmarks data we don't care about any more if necessary).

.github/workflows/collect_benchmarks.yml
Peter554: Hmm it's a good point that the python version might affect performance. Though we don't expect python version to be the primary variable affecting performance, and it's not something we can really influence with our code changes. Therefore I think it's reasonable to just run the benchmarks with a single python version here. I've added the python version to the CSV now though, could be useful information to have.

Peter554: > should we use a matrix and run on multiple python versions?

I think probably not - the benefit seems minimal (seeing the effect of python version on performance) but the cost is significant (considerably slower CI).

.github/workflows/run_benchmarks_main.yml
Peter554: > Can we export this data to an external data store? Like AWS S3 bucket maybe?

We could. I wouldn't rule out that option as something for the future, but it's kind of contradictory to my aim right now. My aim right now is to do something very simple to just get started with collecting this data. If we get AWS involved then this gets considerably more complex.

> Using git for this does not seem ideal because we are not tracking versions for the benchmark files, are we?

I'm not sure I understand your point here. Git seems a quite neat way to store this data to me... 
* The data is kept separate from the main code in a separate branch.
* The data is available by a simple `checkout`. If I want to do some analysis of this data all I need to do is checkout the branch, I don't have to integrate with AWS.

> I think we are also planning to do analysis on these data([comment](https://github.com/octoenergy/xocto/pull/210/files#r1977681870)) using polars . Can we achieve the same results by using a database system? If we are planning to use queries over CSV files using polars, we can do the same with a database system.

We could use a database instead. In the simplest case a SQLite database could replace the CSV file. I'm not sure of the advantage here though. Was your point here just that we could use a database, or that we should use a database?
* Version control will work better with CSVs (clean diffs as we append to the `benchmarks.csv`).
* Querying with polars will be easy and be more than adequate for this data I believe (polars is very fast and powerful).

Peter554: I'm going to resolve this thread, since I think the original reason I opened the thread is resolved. If we want to discuss the other aspects more lets do that in a separate thread.

.github/workflows/run_benchmarks_main.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Setup python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Install dependencies
+        run: make install
+
+      - name: Run benchmarks
+        run: pytest --benchmark-only --benchmark-json benchmarks.json
```
agapius, line 27: Locally, I checked out the main branch of xocto, created a new virtualenv and ran make install.

When I tried to run the benchmarks with `pytest --benchmark-only --benchmark-json benchmarks.json`, I got `pytest: error: unrecognized arguments: --nomigrations --reuse-db --benchmark-only --benchmark-json`

I then checked the [pyproject.toml](https://github.com/octoenergy/xocto/blob/d9872d503ca26732f63b483937dd17c9f0080086/pyproject.toml#L222-L230) file and saw that `--nomigrations` and `--reuse-db` are default options when pytest is run. Both `--benchmark-only` and `--benchmark-json` seem to be pytest-benchmark specifics.
According to gemini, I am "_trying to use Django-specific arguments with the generic pytest command._"

I wonder if we might run into a similar problem during the workflow?

agapius: Resolved - It was a fault on my machine.

.github/workflows/collect_benchmarks.yml
```diff
@@ -0,0 +1,45 @@
...
+          git config user.email github-actions@github.com
+          
+          git add --all
+          git commit -m "Collect benchmarks"
+          
+          # Retry below incase another workflow pushed to the benchmarks-data branch since we checked it out.
+          # Unlikely but technically possible...
+          # There won't be a merge conflict since this workflow is the only workflow touching benchmarks.csv
+          # and this workflow uses `concurrency`.
+          scripts/push_with_retry.sh 5
```
goldenrati0, line 45: cont. from https://github.com/octoenergy/xocto/pull/210#discussion_r1978183146

**tldr;**
For testing purpose this is a good enough setup but for serious benchmarking we *should* use a database system.

---

> If we get AWS involved then this gets considerably more complex.

I would not say considerably more complex, but dependency on the Platform/Infra team
1. Ask Platform/Infra team to create a bucket and provide us `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. _(This step might take a few days to explain the requirements to the team and setup)_
2. Save the values in repository secrets. *(less than 15 minutes)*
3. Update the workflow step to [upload it to S3](https://github.com/marketplace/actions/upload-s3). *(around 3-4 hours with testing and debugging)*

> I'm not sure I understand your point here. Git seems a quite neat way to store this data to me.
> - The data is kept separate from the main code in a separate branch.
> - The data is available by a simple checkout. If I want to do some analysis of this data all I need to do is checkout the branch, I don't have to integrate with AWS.

When you checkout to the `benchmarks-data` you only have the benchmark data and not the codebase. You can run analysis but locally you cannot see the code at the same time, for the code you need to checkout to `main` again. I am not sure but this back-n-forth seems inconvenient. Also, when you are checking benchmark data for a function, you might wanna take a look at the code as well, tweak and test it - with git branching on the same repo, it might not be possible.. Even if we include all the code in the `benchmarks-data` branch, we need to rebase it with `main` and other git related shenanigans periodically.

I am not saying cloud/AWS is our only solution but I feel these data should live elsewhere and not in a different branch in the same repo because it does not scale well. For testing purpose, this is fine I guess but for a more solid approach, I think we should go for a file storage system.

> We could use a database instead. In the simplest case a SQLite database could replace the CSV file. I'm not sure of the advantage here though. Was your point here just that we could use a database, or that we should use a database? My thoughts:
> - Version control will work better with CSVs (clean diffs as we append to the benchmarks.csv).
> - Querying with polars will be easy and be more than adequate for this data I believe (polars is very fast and powerful).

I am trying to say that we *should* use a database, preferably a NoSQL one because I think benchmark data should not be modeled with relationships, hence NoSQL. Additionally, with `polars` we are going to run some queries over the CSV file, which is possible with a database system e.g. NoSQL database like Dynamodb maybe.

In my opinion, some advantages of a database over CSV are
- No need to maintain bulky CSV files
- No need to think about the scale or the amount of data. We can even use the same db system for multiple repositories
- I have not checked `polars` doc extensively but I think we can query the data in a similar pattern without having to involve a file system in our local machine. With our data in cloud, we can connect it to pretty much any applications - integration is easier

Also how often do we want to manually open and check the `benchmarks.csv` file for diffs? If I understood correctly, we are interested in analyzing the data which is possible without opening the file.


Sorry for the long comment 😅 
Let me know if you want to jump on a 🎧 huddle to discuss more.

scripts/extract_benchmarks.py
goldenrati0: Quoting from https://codspeed.io/blog/introducing-codspeed

> this(cloud environments) doesn’t give repeatable measurements either because of the noisy neighbor issue. Basically, physical cloud machines are shared among customers by splitting them into multiple Virtual Machines. At a software level, those are perfectly isolated but in the end, they still share some common hardware and isolation can’t be perfect

By using Github standardised runners, are we not falling into the same issue here? How can we trust the benchmark if it is affected by the noisy neighbors?

_spoiler: Codspeed tries to solve this problem by using micro execution steps. [more](https://codspeed.io/blog/introducing-codspeed)_

agapius (approved):
I think this is super neat! I'd love to be involved in more benchmarking going forward! 🐎 
We might have to adapt our extract-benchmarks.py script ([comment](https://github.com/octoenergy/xocto/pull/210#discussion_r1979611161)) for this to run through - otherwise lets go! 🚢

.github/workflows/collect_benchmarks.yml
Peter554: > When you checkout to the benchmarks-data you only have the benchmark data and not the codebase. ...

This sounds like a reasonable concern. But I think its easily solvable using `git worktree` - you can have two different branches checked out at once.

> I am trying to say that we should use a database. ...

I see your points. I suppose when the benchmarks dataset gets really big then this solution won't work and we'll need another solution (not that likely with xocto any time soon, but would be a problem for Kraken probably). Your suggestion to push files to S3 seems good. We'd then need some kind of ETL to pipe that data through to our chosen datastore. I'll think about it more... (trying to decide if it's worth following up on this now or just start with this simple solution for now)

.github/workflows/run_benchmarks_main.yml
```diff
@@ -0,0 +1,76 @@
+name: Run benchmarks main
+
+on:
+  push:
+    branches:
+      - main
```
goldenrati0, line 6: To test the whole flow, we can add the current branch and remove before merging.

```suggestion
    branches:
      - main
      # todo: remove the branch after testing
      - start-storing-benchmarks-data
```

goldenrati0 (approved):
🙌 Looks good to me for xocto since this is a fairly small repo 😄 
Just left a minor comment to test the flow before merging.

Peter554:
I'm going to hold off on merging this. The approach I've taken here is very lightweight and kind of appealing, but @goldenrati0 makes some good points about scalability. We can't really keep a massive CSV in version control. For xocto this would likely not be a problem any time soon, but if we want to apply this approach to Kraken it likely wouldn't work.

I've asked platform about getting an S3 bucket to dump benchmark results into. This seems like a decent approach. I'm a bit concerned about what would come after that - we'd need some ETL process and a database (something suited to analytics) to load the benchmarks data into. TBH this might all be a bit much. I'm not so worried about setting this thing up - it's some work but I think we could do it. I'm more worried about the work to maintain it. I think it's not really the charges and costs teams job to build (and certainly to _maintain_) this kind of system. 

If we're looking at building something substantial in house, then we should have a team responsible for that system, in order for it to deliver value over the long run. At this point we again might ask if we should be using a 3rd party service e.g. codspeed or bencher. Probably this is worth revisiting since this would be a better solution, if they weren't so expensive.

For now I'm going to put the PR back to draft.

scripts/extract_benchmarks.py
Peter554: Yes you are correct - the noisy neighbour thing can be a problem. I think it's not a big problem for us though:

* pytest-benchmark already runs many rounds/iterations, which reduces the noise quite a lot.
* I think the noise is only a showstopper if you need to measurement very small performance changes. That codspeed article talks detecting 1% performance changes. I am not really looking to catch regressions of that magnitude. I'm more interested in ~10% performance changes.

.github/workflows/collect_benchmarks.yml
agapius: What stops us from starting with this csv approach so see its merit? 
The way I see it, at this moment we just have a few interesting functions for which we want to keep only few records. 
The minimal solution would just keep 2 records: Last and current - this would allow us to detect jumps.
Its also easy to remove again, if we don't see the value.

goldenrati0: I agree with @agapius - this small setup is neat and has the potential to be grown into a full-fledged benchmark CI gradually. Even if we do not merge, we can maybe run the CI in this branch to see how the whole setup works and how the result look like.

Peter554: x-post https://github.com/octoenergy/xocto/pull/210#issuecomment-2703631153

PR #223, opened 2025-08-04
Add timeout and retry options to S3 client

dooferlad:
Prior to this change the S3 client was being created with default options. Sometimes it is desirable to be able to control the retry and timeouts, so this change introduces those options.


xocto/storage/storage.py
```diff
@@ -808,10 +809,22 @@ def _get_policy(self) -> str | None:
         return None
 
     def _get_boto_client(self) -> S3Client:
+        config = {}
+
+        if settings.BOTO_S3_CONNECT_TIMEOUT:
+            config["connect_timeout"] = settings.BOTO_S3_CONNECT_TIMEOUT
+        if settings.BOTO_S3_READ_TIMEOUT:
+            config["read_timeout"] = settings.BOTO_S3_READ_TIMEOU
```
DomMartin27, line 817: 🐼 typo
```suggestion
            config["read_timeout"] = settings.BOTO_S3_READ_TIMEOUT
```

DomMartin27 (approved):

xocto/storage/storage.py
```diff
@@ -808,10 +809,22 @@ def _get_policy(self) -> str | None:
         return None
 
     def _get_boto_client(self) -> S3Client:
+        config = {}
+
+        if settings.BOTO_S3_CONNECT_TIMEOUT:
+            config["connect_timeout"] = settings.BOTO_S3_CONNECT_TIMEOUT
```
jamesbeith, line 815: Seeing as this is an open-source package, I don't think we can be certain that these settings actually exist in Django's settings conf. Could we do something like this instead?


```suggestion
        if (connect_timeout := getattr(settings, "BOTO_S3_CONNECT_TIMEOUT")) is not None:
            config["connect_timeout"] = connect_timeout
```

jamesbeith (dismissed):

PR #231, opened 2025-08-22
Ensure that django settings are integers when populating boto config

jarlidge:
Prior to this change, we were getting settings from django.conf.settings but we weren't ensuring they are integers.

This raises the risk that boto is given strings when it expects integers.

It is also made much easier to override the boto config that the file store uses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Ensure that boto configuration from django settings are integers [#231](https://github.com/octoenergy/xocto/pull/231)
+
 ## V8.4.0 - 2025-09-02
 
 - Add `DjangoJSONValue` Django type [#232](https://github.com/octoenergy/xocto/pull/232)
--- a/xocto/storage/storage.py
+++ b/xocto/storage/storage.py
@@ -127,6 +127,64 @@ class ReadableBinaryFile(Protocol):
     def read(self, size: int = ...) -> bytes: ...
 
 
+def make_boto_config(
+    *,
+    connect_timeout: int | None = None,
+    read_timeout: int | None = None,
+    total_max_attempts: int | None = None,
+) -> botocore.config.Config:
+    """
+    Used to make a botocore.config.Config for use with the S3FileStore class.
+
+    By default that class will use `make_boto_config_from_django_settings` which
+    will populate the inputs to this function using django settings
+    """
+    config: dict[str, object] = {}
+
+    if connect_timeout is not None:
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## V8.3.0 - 2025-08-06
 
 - Add `iterate_over_days` function to `ranges` module [#222](https://github.com/octoenergy/xocto/pull/222)
+- Ensure that boto configuration from django settings are integers [#231](https://github.com/octoenergy/xocto/pull/231)
```
jarlidge, line 8: 8.3.0 has already been released, so this will need to go in a new section

xocto/storage/storage.py
```diff
@@ -126,6 +126,64 @@ class ReadableBinaryFile(Protocol):
...
+
+    def _get_int(key: str) -> int | None:
+        val = getattr(settings, key, None)
+        if val is None or isinstance(val, int):
+            return val
+
+        if isinstance(val, str) and val.isdigit():
+            return int(val)
+
+        raise ValueError(
```
jamesbeith, line 176: Should this be `TypeError`?

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

https://docs.python.org/3/library/exceptions.html#TypeError

xocto/storage/storage.py
```diff
@@ -414,6 +472,11 @@ def _build_download_filepath(self, key_path: str) -> str:
 
 
 class S3FileStore(BaseS3FileStore):
+    """
+    To customise the boto config used and/or the django settings used to populate values in the config,
+    create a subclass that overrides `_get_boto_config`.
```
jamesbeith, line 477: If we're allowing customisation, should the method be public instead of private? I guess this could conflict as to whether we'd want this method to be called externally.

jamesbeith (approved):

CHANGELOG.md
delfick: fair

delfick: right, I'm unsure why I thought that was unreleased. Silly me

delfick: I see what happened, I got confused by the "Unreleased heading" above it. It's fixed now

PR #243, opened 2025-09-18
Remove dependency from ranges module on localtime module

leamingrad:
# Why is this change needed?
Prior to this change, the ranges module had a dependency on the
localtime module, which was only used for a single function call to
localtime.as_localtime that wraps django.utils.timezone.localtime.

# How does it address the issue?
This change removes the dependency. This should make it possible to
import xocto.ranges locally without Django running (which will speed up
some local testing).

### DIFF
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -25,8 +25,6 @@ from typing_extensions import deprecated
 
 from xocto.types import generic
 
-from . import localtime
-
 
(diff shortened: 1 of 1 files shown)


delfick (approved):

jarshwah (approved):
Thanks, had meant to get to this when I saw a dependency on django had been re-introduced here.

PR #265, opened 2026-04-16
Contracts: add ensure_midnight_aligned helper to localtime

andresmtkraken:
# :rocket: Description

### :information_desk_person: Why is this change needed?

As part of reworking datetime usages across the Contract framework, we need a utility to normalise dates and datetimes to localtime midnight-aligned, timezone-aware datetimes at the GraphQL interface boundary. This adds `ensure_midnight_aligned()` to `xocto.localtime`, making it available as a shared utility for Kraken-wide use rather than a one-off in a single domain.

### :sparkles: Type of change

- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/xocto/localtime.py
+++ b/xocto/localtime.py
@@ -657,6 +657,18 @@ def is_aligned_to_midnight(
     )
 
 
+def ensure_midnight_aligned(
+    value: datetime_.date | datetime_.datetime | None,
+) -> datetime_.datetime | None:
+    """
+    Return a localtime midnight-aligned, timezone-aware datetime for the
+    given value, or None if value is None.
+    """
+    if value is None:
+        return None
+    return midnight(value)
+
+
 def consolidate_into_intervals(
     dates: Sequence[datetime_.date],
 ) -> Sequence[Tuple[datetime_.date, datetime_.date]]:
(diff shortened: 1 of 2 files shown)


xocto/localtime.py
```diff
@@ -657,6 +657,28 @@ def is_aligned_to_midnight(
     )
 
 
+def ensure_midnight_aligned(
+    value: datetime_.date | datetime_.datetime,
+) -> datetime_.datetime:
```
andresmtkraken, line 662: Although it will add more checks, I’ve removed the possibility of receiving a `None` value

yoshiohasegawa: Wait a second.. How is this different from `def midnight()`? 🤔

aleksandrfomin-kraken (approved):

xocto/localtime.py
andresmtkraken: Ogg you're right 🤦
I'll allow `None` and just use the `midnight` function

xocto/localtime.py
```diff
@@ -657,6 +657,18 @@ def is_aligned_to_midnight(
     )
 
 
+def ensure_midnight_aligned(
```
yoshiohasegawa, line 660: Hmm, I guess my comment was alluding to the fact that, we probably shouldn't add this if we already have a function that does exactly what we need 🤔

yoshiohasegawa: If all we're gaining is:

```python
    if value is None:
        return None
```

I'm not sure an entirely new function is worth it

PR #278, opened 2026-09-07
Validate finite range endpoints

allanlewis:
In order to improve robustness, this PR forbids setting the bounds of existing ranges.

### DIFF
--- a/tests/fields/postgres/test_ranges.py
+++ b/tests/fields/postgres/test_ranges.py
@@ -469,11 +469,11 @@ class TestHalfFiniteDateTimeRangeField:
             half_finite_datetime_range=half_finite_datetime_range_melb,
             half_finite_datetime_range_utc=half_finite_datetime_range_melb,
         )
-        half_finite_datetime_range_london = ranges.FiniteDatetimeRange(
+        half_finite_datetime_range_london = ranges.HalfFiniteDatetimeRange(
             start=localtime.as_localtime(half_finite_datetime_range_melb.start),
             end=None,
         )
-        half_finite_datetime_range_utc = ranges.FiniteDatetimeRange(
+        half_finite_datetime_range_utc = ranges.HalfFiniteDatetimeRange(
             start=localtime.as_utc(half_finite_datetime_range_melb.start),
             end=None,
         )
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -293,7 +293,7 @@ class Range(Generic[T]):
         return f"<Range: {str(self)}>"
 
     def __setattr__(self, name: str, value: Any) -> None:
-        if name in type(self).__slots__:
+        if any(name in getattr(cls, "__slots__", ()) for cls in type(self).__mro__):
             raise AttributeError("Can't set attributes")
         else:
             super().__setattr__(name, value)
--- a/xocto/ranges.py
+++ b/xocto/ranges.py
@@ -558,6 +558,18 @@ class FiniteRange(Range[T]):
     _start_normalised: T
     _end_normalised: T
 
+    def __init__(
+        self,
+        start: T,
+        end: T,
+        *,
+        boundaries: Union[str, RangeBoundaries] = RangeBoundaries.INCLUSIVE_EXCLUSIVE,
+    ):
+        if any(value is None for value in (start, end)):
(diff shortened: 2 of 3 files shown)


xocto/ranges.py
```diff
@@ -558,6 +558,18 @@ class FiniteRange(Range[T]):
...
+        self,
+        start: T,
+        end: T,
+        *,
+        boundaries: Union[str, RangeBoundaries] = RangeBoundaries.INCLUSIVE_EXCLUSIVE,
+    ):
+        if any(value is None for value in (start, end)):
+            raise ValueError("FiniteRange endpoints cannot be None")
+
+        super().__init__(start, end, boundaries=boundaries)
```
Peter554, line 571: What motivated this change? I'm just wondering if the runtime check is needed? `T` is `Comparable` so the type checker should already reject `None`? At least in a context like Kraken where we are running a type checker.

Not against the change - it makes sense - but a bit more context about why would be appreciated.

xocto/ranges.py
```diff
@@ -293,7 +293,7 @@ def __repr__(self) -> str:
         return f"<Range: {str(self)}>"
 
     def __setattr__(self, name: str, value: Any) -> None:
-        if name in type(self).__slots__:
+        if any(name in getattr(cls, "__slots__", ()) for cls in type(self).__mro__):
```
Peter554, line 296: I assume the `if name in type(self).__slots__:` wasn't working? Can you give a bit more context in the commit message or here in the thread?

xocto/ranges.py
allanlewis: We have a case where we want to backfill historical ranges and a colleague found a hack that you could set the end date of a `HalfFiniteDateTimeRange`, which I don't think should be allowed. The intention seems to be to forbid this but subclasses don't inherit the guard.

xocto/ranges.py
allanlewis: I'm happy to add to the commit message, but the problem is that this isn't being enforced on subclasses like `HalfFiniteDateTimeRange` because it's only checking `cls`.

xocto/ranges.py
Peter554: My question about was about the commit "Validate finite range endpoints". Your response sounds like you're talking about the other commit "Prevent mutation of range endpoints".

allanlewis: Ah, you're right - I got my wires crossed with the mutation commit.

The motivation here is mostly that `FiniteRange` currently lets you construct a value whose `start` or `end` violates its own advertised type. I agree that a type checker should catch this, and I don't have a concrete bug caused by it, I just thought it was useful for the runtime value to maintain the same invariant.

Perhaps this is unnecessary defensive programming though. Happy to drop the commit if you think it is just noise.

xocto/ranges.py
Peter554: Okay - got it thanks

xocto/ranges.py
Peter554: Happy to keep the commit. My only thought really was that it is a (very small) runtime performance hit. In places like costing where we instantiate a lot of ranges this might make a small difference.

Peter554 (approved):
This looks good to me, and should improve correctness.

tests/fields/postgres/test_ranges.py
```diff
@@ -469,11 +469,11 @@ def test_timezone_conversions(self):
             half_finite_datetime_range=half_finite_datetime_range_melb,
```
Peter554, line 469: In the past I've added unreleased changelog notes to the changelog under a "Unreleased" heading. Would it make sense to do that here?

allanlewis: I don't know, you tell me 😄 

Feel free to add that.

FWIW the [development docs](https://xocto.readthedocs.io/en/latest/xocto/development.html) don't mention that, and they're broken in a few places; see my other PRs. (Also see my yet-to-be-created PRs if you have a time machine 😛 )

Peter554: Okay fair. Then I'd say yes, please add to the changelog under the "Unreleased" section. I'd suggest to modify the changelog if the same commit as makes the code change.

PR #282, opened 2026-09-14
Add uv lock file

allanlewis:
In order to improve reproducibility with respect to CI, this commit adds a uv
lock file. A pre-commit hook and a CI step check that the lockfile is
up-to-date. The lock file is treated as binary in order that its often large
diffs don't appear in `git show` etc. or GitHub's diff view without
specifically requesting that, and also to discourage manual merging of changes.


ColinBe95 (approved):
looks good

PR #283, opened 2026-09-14
Use PEP 735 dependency groups

allanlewis:
In order to improve developer experience, this commit switches the
non-production dependencies from "extras" to "dependency groups", as defined by
[PEP 735].

The intent of 'extras' is to provide optional dependencies for consumers of the
package. Development and documentation dependencies aren't that category,
they're purely for those developing the library.

[PEP 735]: https://peps.python.org/pep-0735/


ColinBe95 (approved):
makes sense, lgtm!

PR #289, opened 2026-09-17
Upgrade CI Postgres to support newer versions of Django

allanlewis:
In order to allow CI to succeed with newer versions of Django, this commit bumps PostgreSQL to v15.


.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ jobs:
   build:
     docker:
       - image: cimg/python:3.10
-      - image: cimg/postgres:14.20
+      - image: cimg/postgres:15.15
```
jamesbeith: How come you landed on `15.15` and not `15.19` that appears to be [available](https://hub.docker.com/layers/cimg/postgres/15.19/images/sha256-08d4cc58f58ef41fd9d20c05fd5f3184613a25c1f1a51e98ec26969f4cb75231)?

jamesbeith (approved):

.circleci/config.yml
allanlewis: This was part of a larger branch and that's the version the LLM picked 🤖 

I'll update it to 15.19 before I merge.

PR #291, opened 2026-09-18
Replace mypy-boto3 packages with boto3-stubs

allanlewis:
## Summary

- replace the service-specific Mypy package with `boto3-stubs`
- apply generated S3 request, literal, and blob types to the storage module
- retain Python 3.10-compatible runtime handling for S3 blob values
- run Mypy with the locked uv environment

## Testing

```sh
make ci
uv run prek run --all-files
uv lock --check
```

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


allanlewis:
Superseded by the rewritten Mypy stack in #294 and #290.

PR #295, opened 2026-09-18
Use uv-managed environments in CI

allanlewis:
As per the sole commit:

> In order to run project checks consistently, this commit uses uv-managed tools locally and uv-managed dependencies in CircleCI.


leamingrad (approved):
LGTM!

PR #297, opened 2026-09-18
Tidy `pyproject.toml`

allanlewis:
In order to reduce complexity, this commit tidies `pyproject.toml`, removing redundant configuration and fixing broken links.


pyproject.toml
```diff
@@ -215,15 +178,9 @@ section-order = [
 #   Hence the rules ignoring deprecation warnings must by below the rule that converts
 #   DeprecationWarnings into errors.
 filterwarnings = [
-    "error::ResourceWarning",
-    "error::RuntimeWarning",
-    "error::DeprecationWarning",
-    "error::pytest.PytestCollectionWarning",
+    "error",
```
jamesbeith, line 181: Could you explain this change? Does a sole `error` turn all warnings into errors?

jamesbeith (approved):

pyproject.toml
allanlewis: Yes, precisely that 😄

